PR #31, opened 2025-01-23
Feature Flag Local Evaluation Implementation

### LINKED ISSUE #3: Implement Local Evaluation for Feature Flags
Currently, the client makes a request to `/decide` for every feature flag evaluation. The plan is to add an option to allow the client to perform local evaluation when the personal API key is supplied.

haacked:
Opening this as a draft in case anyone wants to follow along and check my work. This one requires a lot of attention to detail.

Fixes #3 

## Todo

- [x] Implement an authenticated request to `/feature_flag/local_evaluation`.
- [x] Implement a type to deserialize the `/feature_flag/local_evaluation` endpoint to.
- [x] Review `LocalEvaluationApiResult` and make sure it can handle anything [returned by the endpoint](https://github.com/PostHog/posthog/blob/fix/filter-behavioral-cohorts-in-ff-ui/posthog/api/feature_flag.py#L695-L777).
- [x] Make a decision on how to handle `Values`. So far I've seen that it can be a string, string array, int, bool. Currently using a `JsonElement`, but if that's the full set, I may create a specific type similar to [the approach I took with StringOrValue&lt;T>](https://haacked.com/archive/2024/12/17/string-or-value/).
- [x] Implement the actual evaluation using the [posthog-python implementation](https://github.com/PostHog/posthog-python/blob/master/posthog/feature_flags.py) as a guide.
- [x] Implement caching the response and polling for results
- [x] Try and create every possible type of feature flag and test it.
- [x] Add unit tests


haacked:
Local Evaluation is complete. Ported pretty much all the tests from https://github.com/PostHog/posthog-python for local evaluation.

PR #43, opened 2025-02-06
fix: remove extra closing brace in capture call

lurumad:
Removed an unnecessary closing brace in the posthogClient.Capture call, ensuring correct syntax.


haacked:
Thanks @lurumad!

PR #46, opened 2025-02-11
Stop adding local feature flag data to $feature_flag_called event

haacked:
See https://github.com/PostHog/posthog-python/pull/181 for more context.

Also, remove is_simple_flag from unit tests. It's not used anywhere.


haacked:
Ugh, it's failing `prettier --write "./**/*.{ts,tsx}"` but that command passes on my local machine. Also, it's failing on a file I didn't change in my PR.

PR #56, opened 2025-02-21
feat(feature-flags): support quota limiting for feature flags #403

dmarticus:
with https://github.com/PostHog/posthog/pull/28564, we now start to respond different with the /decide and /local_evaluation APIs if users have gone over their quota limit. Now we need to change the SDKs to handle these new responses.

### DIFF
--- a/src/PostHog/Api/DecideApiResult.cs
+++ b/src/PostHog/Api/DecideApiResult.cs
@@ -14,7 +14,8 @@ internal record DecideApiResult(
     Analytics? Analytics = null,
     bool DefaultIdentifiedOnly = true,
     bool ErrorsWhileComputingFlags = false,
-    IReadOnlyDictionary<string, string>? FeatureFlagPayloads = null);
+    IReadOnlyDictionary<string, string>? FeatureFlagPayloads = null,
+    IReadOnlyList<string>? QuotaLimited = null);
 
 internal record FeatureFlagsConfig([property: JsonPropertyName("enable_collect_everything")] bool EnableCollectEverything);
 internal record Analytics(string Endpoint);
--- a/src/PostHog/Api/DecideApiResult.cs
+++ b/src/PostHog/Api/DecideApiResult.cs
@@ -7,9 +7,6 @@ namespace PostHog.Api;
 /// The API Result for the <c>/decide</c> endpoint. Some properties are ommitted because
 /// they are not necessary for server-side scenarios.
 /// </summary>
-/// <remarks>
-///
-/// </remarks>
 internal record DecideApiResult(
     FeatureFlagsConfig? Config = null,
     bool IsAuthenticated = false,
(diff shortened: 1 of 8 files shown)


haacked:
Thanks @dmarticus! I'll fix this up.

dmarticus:
@haacked thanks!  I was working on it locally but my .NET LSP isn't playing nicely so it's been a pain.  Thanks for helping, lmk if you have questions about the implementation!

src/PostHog/Api/DecideApiResult.cs
```diff
@@ -17,7 +17,8 @@ internal record DecideApiResult(
     Analytics? Analytics = null,
     bool DefaultIdentifiedOnly = true,
     bool ErrorsWhileComputingFlags = false,
-    IReadOnlyDictionary<string, string>? FeatureFlagPayloads = null);
+    IReadOnlyDictionary<string, string>? FeatureFlagPayloads = null,
+    IReadOnlyList<string>? QuotaLimited = null);
```
haacked, line 21: If I understand the unit tests, `QuotaLimited` isn't ever going to be a part of the successful `/decide` response body. Instead, it will be an error response with the status `HttpStatusCode.PaymentRequired` and a response body that looks like this:

```json
{
    "type": "quota_limited",
    "detail": "You have exceeded your feature flag request quota",
    "code": "payment_required"
} 
```

Correct?

haacked:
When attempting a local evaluation API call that is quota limited, I assume we don't want to fallback to `/decide` correct? If one is quota limited, do we assume the other is too? Or should we continue to fallback to `/decide`?

dmarticus:
> When attempting a local evaluation API call that is quota limited, I assume we don't want to fallback to /decide correct? If one is quota limited, do we assume the other is too? Or should we continue to fallback to /decide?

Yup, we should short-circuit if we get a local eval call that's quota limited

dmarticus:
> @dmarticus I also noticed the log level for quota exceeded is Debug in this PR. I'm thinking this is something a consumer would want to notice. It should probably be Warning or Error. Right?

Yeah, let's go with a warning log, that's what the other PRs do.

src/PostHog/Api/DecideApiResult.cs
haacked: Ah, thanks for the clarification! I'll fix things back up. Do you have an example `/decide` response JSON with the quota info?

dmarticus: I mean it's just like another `/decide` response, only with empty flags and payloads and this new field.  Looks like this
```
{
  "config": {
    "enable_collect_everything": true
  },
  "toolbarParams": {},
  "isAuthenticated": false,
  "supportedCompression": [
    "gzip",
    "gzip-js"
  ],
  "featureFlags": {},
  "featureFlagPayloads": {},
  "errorsComputingFlags": false,
  "quotaLimited": ["feature_flags"] // it's like this in case we have other quota-limited resources, e.g. "recordings"
  "sessionRecording": false
}
```

haacked (approved):
Nice work!

PR #60, opened 2025-03-05
Split IPostHogClient so base interface only contains public methods.

bartsaintgermain:
`IPostHogClient` contains an internal interface. This complicates unit testing projects which include this library.

This MR does split the `IPostHogClient` where 
- `IPostHogBaseClient` only contains public methods
- `IPostHogClient` inherits `IPostHogBaseClient` and the internal method.


haacked:
Hi @bartsaintgermain! Thanks for the PR. I did feel dirty adding that internal method. However, I don't want to expose that method at all. It's only there to support ASP.NET Core feature management. I'm going remove the method and come up with a different way to support feature management.

bartsaintgermain:
thx @haacked,

PR #84, opened 2025-06-26
feat: use .NET 8 packages

Thanaen:
This PR updates the package dependencies to match .NET 8 versions instead of .NET 9 versions.

## Motivation
I've tried using this library in my .NET 8 project, but updating all those libs to the .NET 9 version broke my app. The library was previously using .NET 9 package versions, which made it difficult to use in .NET 8 projects. By aligning the package versions with .NET 8, this change makes the library easier to integrate and use in .NET 8 projects while maintaining compatibility.

## Changes
- Updated package dependencies from .NET 9 versions to .NET 8 versions

## Benefits
- Improved compatibility with .NET 8 projects
- Easier integration for developers using .NET 8
- Maintains functionality while using appropriate package versions

## Testing
- [ ] Existing tests pass
- [ ] No breaking changes introduced
- [ ] Verified compatibility with .NET 8 projects

### DIFF
--- a/src/PostHog/Api/ComparisonOperator.cs
+++ b/src/PostHog/Api/ComparisonOperator.cs
@@ -11,90 +11,90 @@ public enum ComparisonOperator
     /// <summary>
     /// Matches if the value is in the list of filter values. Only used for cohort filters.
     /// </summary>
-    [JsonStringEnumMemberName("in")]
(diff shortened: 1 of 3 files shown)


haacked:
I’m looking into it

PR #97, opened 2025-08-22
feat(flags): Implement local evaluation of flag dependencies

haacked:
This implements flag dependency evaluation for the .NET SDK [following the same patterns as the Python SDK](https://github.com/PostHog/posthog-python/pull/311). Flag dependencies allow flags to depend on the evaluation result of other flags.


andyzzhao:
LGTM. It would be nice to have a test that has a flag dependency and another non flag dependency property match.

PR #101, opened 2025-08-26
feat: allow capture to override timestamp with datetime or datetimeoffset

darkopia:
my previous attempt changed the wrong place but should likely remain.

I am trying to add timestamp override support to .Net from [this support ticket.](https://posthoghelp.zendesk.com/agent/tickets/36274)

Other SDK's support sending timestamps as a property instead of using current datetime.


haacked:
Rather than try to extract `timestamp` from properties, we should add one or more extension method overrides of `Capture` to `src/PostHog/Capture/CaptureExtensions.cs` that takes in a `timestamp` and passes it along to the properties.

PR #105, opened 2025-09-13
feat: Add optional distinctId to GroupIdentifyAsync

luke-belton:
Currently we set the `distinctId` on `GroupIdentifyAsync` calls using our own generated identifier based on the format `${group_type}_{group_key}`. While this works, it's out of keeping with other SDKs where generally we also allow users to set their own `distinctId` (e.g [Python](https://github.com/PostHog/posthog-python/blob/337f7da7c585a8fa04da8945c1fe6beccb05c6a5/posthog/client.py#L343-L346)). While it's not a major issue as groups are technically used for aggregating events (not persons) and therefore the `$groupidentify` event can just be used to announce the group to PostHog and add/update properties, i.e. not necessarily needing to be linked to a specific person via `distinctId`. That said - it's still a bit of a jarring experience and it would bring the .NET package in line with other SDKs if we allowed it to be set.

I've made `distinctId` optional to avoid making a breaking change.

User request context: https://posthoghelp.zendesk.com/agent/tickets/37401. (moved to PostHog: https://us.posthog.com/project/2/support/tickets/41560)

Tested the changes locally in the sample app and they seem to work as expected ([event](https://us.posthog.com/project/35787/events/019970f1-7fef-7b20-a6cb-c76c5627b9b3/2025-09-22T12%3A21%3A39.024000%2B02%3A00)).

### DIFF
--- a/src/PostHog/Api/PostHogApiClientExtensions.cs
+++ b/src/PostHog/Api/PostHogApiClientExtensions.cs
@@ -52,10 +52,11 @@ internal static class PostHogApiClientExtensions
         string type,
         StringOrValue<int> key,
         Dictionary<string, object>? properties,
-        CancellationToken cancellationToken)
+        CancellationToken cancellationToken,
+        string? distinctId = null)
     {
         return await client.SendEventAsync(
-            distinctId: $"${type}_{key}",
+            distinctId: distinctId ?? $"${type}_{key}",
             eventName: "$groupidentify",
             properties: new Dictionary<string, object>
             {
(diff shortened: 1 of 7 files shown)


luke-belton:
@haacked wondered if you'd be able to take a look at this? Is it something we'd consider adding?

src/PostHog/Api/PostHogApiClientExtensions.cs
```diff
@@ -52,10 +52,11 @@ public static async Task<ApiResult> GroupIdentifyAsync(
         string type,
         StringOrValue<int> key,
         Dictionary<string, object>? properties,
-        CancellationToken cancellationToken)
+        CancellationToken cancellationToken,
+        string? distinctId = null)
```
haacked, line 56: For public API's, I prefer to avoid optional parameters like this because it can cause [subtle versioning issues down the road](https://haacked.com/archive/2010/08/12/more-optional-versioning-fun.aspx/).

Instead, consider adding an overload with distinctId.

haacked (changes requested):
Thanks! I'll take this PR after the requested change. Sorry I didn't get to it earlier.

haacked (approved):
Nicely done. Thanks!

Issue #33, opened 2025-02-03
Implement capturing exceptions


haacked:
The `posthog-python` library supports methods that capture exception information. It does not look like `posthog-node` supports this. 

Here's the signature:

```python
def capture_exception(
    exception=None,  # type: Optional[BaseException]
    distinct_id=None,  # type: Optional[str]
    properties=None,  # type: Optional[Dict]
    context=None,  # type: Optional[Dict]
    timestamp=None,  # type: Optional[datetime.datetime]
    uuid=None,  # type: Optional[str]
    groups=None,  # type: Optional[Dict]
):
```

And here's the docs in the code:

```python
"""
    capture_exception allows you to capture exceptions that happen in your code. This is useful for debugging and understanding what errors your users are encountering.
    This function never raises an exception, even if it fails to send the event.

    A `capture_exception` call does not require any fields, but we recommend sending:
    - `distinct id` which uniquely identifies your user for which this exception happens
    - `exception` to specify the exception to capture. If not provided, the current exception is captured via `sys.exc_info()`

    Optionally you can submit
    - `properties`, which can be a dict with any information you'd like to add
    - `groups`, which is a dict of group type -> group key mappings

    For example:
    ```python
    try:
        1 / 0
    except Exception as e:
        posthog.capture_exception(e, 'my specific distinct id')
        posthog.capture_exception(distinct_id='my specific distinct id')

    ```
    """
```

We may want to implement a .NET version of this.

darielkremov:
Following up on the discussion in #108 - I would be happy to help with the automatic exception capture module, but I think it's worth discussing the spec in some more detail first. Would be great if we also get some input from folks that have implemented the autocapture in the other libraries so we get a consistent behavior, especially for edge cases that have already been thought of before.

Some initial thoughts:
- Need to ensure we only count an exception once
- Should have rate limits & retries on submitting exception events (implement #7 first?)
- Options to control enabled/disabled and queue size 
- Should queue exceptions up to some limit, support periods of offline usage, etc. Queue/flush options for events in general will already work and handle this, but counting exceptions separately is worth considering.
[...]

PR #108, opened 2025-09-25
feat(err): Manual capture of exceptions

darielkremov:
**TL;DR** - This PR allows library users to manually capture exception events and view them in PostHog, using `CaptureException()`.

Wanted to play around a bit, and as autocapture would be a lot more work, thought manual capture could be useful anyway, say first stage of implementing #33.
I've tried to follow and be as consistent as possible with the `posthog-python` and `posthog-js` implementations, though they do differ a bit. Ideally, I would have preferred to refactor a bit to use types instead of the dictionaries when building out the properties, but decided against to minimize changes for now.

Some more things to note:
- `$exception_personURL` actually works in the PostHog dashboard, even though I'm not a huge fan of how it gets constructed
- `["platform"] = "custom"` and `["lang"] = "dotnet"` so cymbal doesn't go crazy with exceptions trying to figure out what's happening - I assume `platform` would have to be changed when officially supported. Either way, context actually works on client code, of course with the obvious caveats that you need the .pdb debug symbols, system libraries give empty context etc.
- OS name and version detection is more accurate for Windows machines, but likely to be inconsistent with other client libraries (e.g. `posthog-python` would detect Windows 11 as Windows 10, here we would have the correct version)
- Tests for the exception event properties don't use the entire JSON as it would be too much and, of course, would break on paths, etc.
- Added a button to the sample HogTied webapp to trigger an exception event for an easy tryout as well. However, I noticed the webapp code seems to be missing a dependency (bootstrap) and might look "broken". If anyone struggles with that -> Add Client Side Library -> install bootstrap and double check paths in `Pages/Shared/_Layout.cshtml`. Didn't want to clutter the PR with that.
- Basically all constants taken as they are in `posthog-python` (max length, lines, etc)

### DIFF
--- a/src/PostHog/Library/StringExtensions.cs
+++ b/src/PostHog/Library/StringExtensions.cs
@@ -10,7 +10,7 @@ internal static class StringExtensions
     /// Truncates <paramref name="value"/> so its UTF-8 byte length is at most <paramref name="maxLength"/>.
     /// Appends <paramref name="truncationSymbol"/> if truncation occurs. Respects UTF-8 code point boundaries, but it can split graphemes.
     /// </summary>
-    /// <param> name="value">The string to truncate.</param>
+    /// <param name="value">The string to truncate.</param>
     /// <param name="maxLength">Maximum allowed byte length in UTF-8 encoding.</param>
     /// <param name="truncationSymbol">Optional: The symbol to append if truncation occurs. Default is "…".</param>
     /// <returns>The truncated string, if longer than <paramref name="maxLength"/>, otherwise original <paramref name="value"/>.</returns>
@@ -48,7 +48,7 @@ internal static class StringExtensions
     /// Truncates <paramref name="value"/> so it is at most <paramref name="maxLength"/> characters.
     /// Appends <paramref name="truncationSymbol"/> if truncation occurs.
     /// </summary>
-    /// <param> name="value">The string to truncate.</param>
+    /// <param name="value">The string to truncate.</param>
     /// <param name="maxLength">Maximum allowed characters.</param>
     /// <param name="truncationSymbol">Optional: The symbol to append if truncation occurs. Default is "…".</param>
     /// <returns>The truncated string, if longer than <paramref name="maxLength"/>, otherwise original <paramref name="value"/>.</returns>


src/PostHog/ErrorTracking/ExceptionPropertiesBuilder.cs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private static SourceCodeContext BuildSourceCodeContext(
+        string absolutePath,
+        int lineNumber,
+        int columnNumber,
+        int maxLines = 0,
+        int maxLength = 0)
+    {
+        var lines = File.ReadAllLines(absolutePath);
```
haacked, line 129: This line could throw an exception for a variety of reasons. I recommend adding a method for reading the path safely and using that:

```csharp
private static string? ReadSourceCodeSafely(string absolutePath)
{
    if (string.IsNullOrEmpty(absolutePath) || !File.Exists(absolutePath))
        return null;

    try
    {
        return File.ReadAllLines(absolutePath);
    }
    catch (Exception ex) when (
        ex is IOException or UnauthorizedAccessException or NotSupportedException)
    {
        return null;
    }
}
```

src/PostHog/ErrorTracking/ExceptionPropertiesBuilder.cs
```diff
@@ -0,0 +1,145 @@
...
+
+    private static List<Dictionary<string, object>> BuildExceptionList(Exception exception)
+    {
+        var list = new List<Dictionary<string, object>>();
+        var seen = new HashSet<Exception>();
+
+        var stack = new Stack<Exception>();
+        stack.Push(exception);
+
+        while (stack.Count > 0)
```
haacked, line 39: No depth limit could result in a very large collection in pathological cases. Consider adding a depth limit and max exception count.

src/PostHog/Library/OperatingSystemInfo.cs
```diff
@@ -0,0 +1,129 @@
+﻿using System.Runtime.InteropServices;
+
+namespace PostHog.Library;
+
+/// <summary>
+/// Provides human friendly version information about the current operating system.
+/// </summary>
+public static class OperatingSystemInfo
```
haacked, line 8: I'd prefer the following pattern to make the immutability more explicit:

```csharp
  public static class OperatingSystemInfo
  {
      private static readonly string _name;
      private static readonly string _version;

      public static string Name => _name;
      public static string Version => _version;

      static OperatingSystemInfo()
      {
          // Initialize _name and _version here
          // This makes immutability more explicit
      }
  }
```

darielkremov: Sure. Will actually get rid of the explicit static constructor altogether to avoid CA1810.

src/PostHog/ErrorTracking/ExceptionPropertiesBuilder.cs
darielkremov: Found some default depth & frame limits from the js client, will set the same limits here. And a max exceptions one that is basically just to cap unhinged AggregateExceptions.

src/PostHog/ErrorTracking/ExceptionPropertiesBuilder.cs
```diff
@@ -0,0 +1,202 @@
...
+            }
+
+            var lineIndex = lineNumber - 1;
+
+            int start = Math.Max(0, lineIndex - maxLines);
+            int end = Math.Min(lines.Length - 1, lineIndex + maxLines);
+            var truncatedLines = lines.Select(l => l.Trim('\r', '\n')
+                .TruncateByBytes(maxLength)
+                .TruncateByCharacters(maxLength))
+                .ToArray();
```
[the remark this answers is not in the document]
haacked: Hmm, well that's disappointing that it wasn't fixed. In any case, is everything else done with this PR as far as you're concerned?

src/PostHog/Library/StringExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        string head = utf8.GetString(bytes, 0, cut);
+        return head + truncationSymbol;
+    }
+
+    /// <summary>
+    /// Truncates <paramref name="value"/> so it is at most <paramref name="maxLength"/> characters.
+    /// Appends <paramref name="truncationSymbol"/> if truncation occurs.
+    /// </summary>
+    /// <param> name="value">The string to truncate.</param>
```
haacked:
```suggestion
    /// <param name="value">The string to truncate.</param>
```

PR #118, opened 2025-10-17
Fix feature flag cache key to include person properties and groups

### LINKED ISSUE #75: Not enough data to generate correct key for caching
To enable caching for the 'decide' endpoint, you can implement the IFeatureFlagCache interface. However, there's a caveat: the `GetAndCacheFeatureFlagsAsync` method only accepts the `distinctId` as a parameter. This is insufficient to build the cache correctly, because the endpoint can return different results for the same distinctId depending on whether PersonalProperties were provided. Therefore, an additional PersonalProperties parameter is needed for accurate caching.

    public async Task<IReadOnlyDictionary<string, FeatureFlag>>
        GetAndCacheFeatureFlagsAsync(
        string distinctId, 
        **Dictionary<string, object?>? PersonProperties**, 
        Func<CancellationToken, Task<IReadOnlyDictionary<string, FeatureFlag>>> fetcher, 

haacked:
The cache key previously only used distinctId, which caused incorrect cached results when the same user had different person properties or belonged to different groups. The `/flags` endpoint can return different feature flags based on these parameters.

Changes:
- Add `FeatureFlagCacheKey` class to generate stable cache keys from distinctId, person properties, and groups
- Add new `IFeatureFlagCache.GetAndCacheFlagsAsync` overload with personProperties and groups parameters
- Mark old `GetAndCacheFlagsAsync` method as obsolete
- Update all built-in cache implementations to use new cache key
- Update `PostHogClient` to pass properties and groups to cache
- Add comprehensive unit and integration tests

The changes are backward compatible. Custom `IFeatureFlagCache` implementations will continue to work with a safe default behavior (no caching) until updated to override the new method.

Fixes #75

### DIFF
--- a/samples/HogTied.Web/Pages/Index.cshtml
+++ b/samples/HogTied.Web/Pages/Index.cshtml
@@ -62,6 +62,51 @@
                 </div>
             }
 
+            @if (Model.CacheDemo is not null && Model.UserId is not null) {
+                <div class="alert alert-success mt-3" role="alert">
+                    <h5 class="alert-heading">🚀 Request-Level Cache Demo</h5>
+                    <p>
+                        To demonstrate that feature flags are cached per HTTP request, we checked the
+                        <code>simple-flag</code> twice during this page load:
+                    </p>
+                    <table class="table table-sm table-bordered bg-white mt-2 mb-2">
+                        <thead>
+                            <tr>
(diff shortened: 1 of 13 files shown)


gustavohstrassburger (approved):
Nice work on the tests.

PR #126, opened 2025-11-20
included evaluated_at properties in $feature_flag_called events

dmarticus:
See https://github.com/PostHog/posthog/pull/41824 for context


src/PostHog/Features/SendFeatureFlagsOptions.cs
```diff
@@ -0,0 +1,29 @@
+namespace PostHog;
+
+/// <summary>
+/// Options to control how feature flags are sent with captured events.
+/// </summary>
+public class SendFeatureFlagsOptions
```
haacked: I don't see where this is used.

PR #138, opened 2025-12-08
Add PostHog.AI Package for LLM Observability

matjanos:
## Summary

Adds a new `PostHog.AI` package to provide AI observability for .NET applications. Intercepts OpenAI API requests and responses to capture events with properties per the PostHog AI event spec.

## Features

- Automatic event capture: captures `$ai_generation` and `$ai_embedding` events from OpenAI API calls
- Model parameters: extracts temperature, max_tokens, stream, tools from requests
- Token usage: captures input/output/total tokens from responses
- Streaming support: handles Server-Sent Events (SSE) with proper message parsing
- Context management: `PostHogAIContext` for session/span tracking and custom properties
- Cache tracking: captures cache token usage when available
- Error handling: captures errors and HTTP status codes

### Core Components

- `PostHogOpenAIHandler`: Delegating handler that intercepts HTTP requests/responses
- `PostHogAIContext`: Async-local context for setting session IDs, span IDs, and custom properties
- `PostHogAIExtensions`: Dependency injection extensions for setup

### Captured Properties

The handler extracts and captures properties including:
- Core: trace_id, provider, model, latency, URLs, HTTP status
- Model parameters: temperature, max_tokens, stream, tools
- Token usage: input_tokens, output_tokens, total_tokens
- Context: session_id, span_id, span_name, parent_id (via PostHogAIContext)
- Cache: cache_read_input_tokens, cache_creation_input_tokens
- Errors: is_error, error messages

### Streaming Support

- Proper SSE message parsing that handles partial messages across chunks
- Buffer management for incomplete lines
- Support for both `\n` and `\r\n` line endings (with `\r\n` checked first)

## Usage Example

```csharp
// Configure PostHog
builder.Services.AddPostHog(options => {
    options.ProjectApiKey = "YOUR_KEY";
    options.HostUrl = new Uri("https://us.i.posthog.com");
});

// Register OpenAI client with PostHog integration
builder.Services.AddPostHogOpenAIClient("YOUR_OPENAI_API_KEY");

// Use with context
using (PostHogAIContext.BeginScope(
    sessionId: "session-123",
    spanId: "span-1",
    spanName: "summarize_text"))
{
    var response = await openAIClient.GetChatCompletionsAsync(...);
}
```

## Testing

- Unit tests for non-streaming requests
- Tests for embedding events
- Tests for streaming responses with SSE parsing
- All tests passing

## Files Changed

- New package: `src/PostHog.AI/PostHog.AI.csproj`
- New tests: `tests/PostHog.AI.Tests/PostHog.AI.Tests.csproj`
- Updated solution file to include new projects
- Updated root README to document the new package
- Fixed typo in `samples/HogTied.Web/Program.cs` (PostHoc → PostHog)

## Notes

- Cost properties (`$ai_input_cost_usd`, etc.) are only set when explicitly provided via `PostHogAIContext.Properties`
- The handler skips null values when merging context properties
- Backward compatible with existing PostHog setup


src/PostHog.AI/PostHogOpenAIHandler.cs
```diff
@@ -0,0 +1,744 @@
...
+
+            // Request extraction
+            string? model = null;
+            if (requestJson != null)
+            {
+                model = requestJson["model"]?.ToString();
+                eventProperties["$ai_model"] = model ?? "";
+
+                // Extract input
+                if (requestJson["messages"] is JsonArray messages)
```
haacked: `$ai_input` and `$ai_output_choices` can contain sensitive information. We should not set these if `PostHogAIContext.PrivacyMode` is set to `true`.

src/PostHog.AI/PostHogOpenAIHandler.cs
```diff
@@ -0,0 +1,744 @@
...
+        string? contentString = null;
+        JsonNode? jsonNode = null;
+
+        try
+        {
+            if (content != null)
+            {
+#pragma warning disable CA2016 // Forward cancellation token (ReadAsStringAsync has an overload with CancellationToken in .NET 5+, LoadIntoBufferAsync doesn't)
+                await content.LoadIntoBufferAsync();
+                contentString = await content.ReadAsStringAsync(cancellationToken);
```
haacked: This buffers the content twice - once in `LoadIntoBufferAsync()` and again when `ReadAsStringAsync()` creates a string. For large responses, consider parsing JSON directly from the stream:

```csharp
using var stream = await content.ReadAsStreamAsync(cancellationToken);
jsonNode = await JsonNode.ParseAsync(stream, cancellationToken:
cancellationToken);
```

src/PostHog.AI/PostHogOpenAIHandler.cs
```diff
@@ -0,0 +1,744 @@
...
+                // Ignore other errors during chunk processing to avoid crashing the stream.
+                // The stream should still flow to the client.
+            }
+#pragma warning restore CA1031
+        }
+
+        private void ProcessSSEMessage(string message)
+        {
+            // Split message into lines
+            var lines = message.Split(LineSeparators, StringSplitOptions.RemoveEmptyEntries);
```
haacked: This allocates a new string array for each SSE message chunk. For streaming responses with 50-100+ chunks, this adds up. Since you're targeting net6.0+, you could use span-based enumeration:

```csharp  
foreach (var line in message.AsSpan().EnumerateLines())
```

src/PostHog.AI/PostHog.AI.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
haacked: Probably should target the same thing that the [core PostHog lib does](https://github.com/PostHog/posthog-dotnet/blob/main/src/PostHog/PostHog.csproj#L4):

```suggestion
    <TargetFrameworks>netstandard2.1;net8.0</TargetFrameworks>
```

haacked (changes requested):
Thanks for submitting this. Added some comments.

src/PostHog.AI/PostHog.AI.csproj
matjanos: totally! fixed

haacked:
@matjanos I'll take another look. Sorry for the delay!

haacked (commented):
Getting real close! I found what looks like a bug. Once that's fixed, I can approve.

Also, consider adding tests for: network/connection errors, HTTP 4xx/5xx responses, malformed JSON handling, and the `AddPostHogOpenAIClient` validation logic.

src/PostHog.AI/PostHogOpenAIHandler.cs
```diff
@@ -0,0 +1,821 @@
...
+        if (request is null)
+        {
+            throw new ArgumentNullException(nameof(request));
+        }
+#endif
+
+        var stopwatch = Stopwatch.StartNew();
+        var requestJson = await ReadContentAndParseJsonAsync(
+            request.Content,
+            ex => _logger.LogRequestContentFailure(ex),
```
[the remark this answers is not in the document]
matjanos: Fixed. Added `await request.Content.LoadIntoBufferAsync()` before parsing, and switched `ReadContentAndParseJsonAsync` to use `ReadAsByteArrayAsync` + `new MemoryStream(bytes)` instead of `ReadAsStreamAsync` so the content is never consumed before `base.SendAsync`.

src/PostHog.AI/PostHogOpenAIHandler.cs
```diff
@@ -0,0 +1,821 @@
...
+        {
+            response = await base.SendAsync(request, cancellationToken);
+        }
+        catch (Exception ex)
+        {
+            stopwatch.Stop();
+            // Capture error event
+            _ = Task.Run(
+                () =>
+                    CaptureEventAsync(
```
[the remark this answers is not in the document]
matjanos: Fixed. Renamed to `CaptureEvent` with `void` return. All 3 call sites now call it inline — no more `Task.Run` wrapping.

PR #142, opened 2025-12-16
fix: Capture doesn't send flags when sendFeatureFlags=false

dustinbyrne:
According to the documentation
https://github.com/PostHog/posthog-dotnet/blob/8648856ebd8dc3cb589d361f3fc0d932aceb3291/src/PostHog/Capture/CaptureExtensions.cs#L36

When `sendFeatureFlags` is `true`, it should be expected that flags properties are appended to the event. When `false`, it should be expected that the opposite occurs - flags properties _are not_ appended.

Related to https://github.com/PostHog/posthog/issues/31425


haacked (approved):
:shipit:

PR #144, opened 2025-12-18
feat: Set `$feature_flag_error` on `$feature_flag_called`

dustinbyrne:
This change adds observability into remote feature flag evaluation failures by including a `$feature_flag_error` property on `$feature_flag_called` events. The error types are:                                 
                                                                                                                                                                                                      
| Error Type                   | Description                                                       |                                                                                                  
|------------------------------|-------------------------------------------------------------------|                                                                                                  
| timeout                      | Request timed out (`TaskCanceledException` when not user-cancelled) |                                                                                                  
| connection_error             | Network connectivity issues (`HttpRequestException`)                |                                                                                                  
| api_error_{status}           | API returned HTTP error with status code (e.g., `api_error_500`)    |                                                                                                  
| errors_while_computing_flags | Server returned `errorsWhileComputingFlags: true`                   |                                                                                                  
| quota_limited                | Feature flags quota exceeded                                      |                                                                                                  
| flag_missing                 | Requested flag not in API response                                |                                                                                                  
| unknown_error                | Unexpected error that doesn't match known patterns                |                                                                                                  
                                                                                                                                                                                                      
Multiple errors can occur simultaneously and are joined with commas (e.g., `errors_while_computing_flags,flag_missing`).

### DIFF
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -403,7 +403,7 @@ public sealed class PostHogClient : IPostHogClient
             {
                 HandleRemoteError(e, FeatureFlagError.ApiError((int)e.Status));
             }
-            catch (Exception e) when (e is not ArgumentException and not NullReferenceException)
+            catch (Exception e) when (e is not ArgumentException and not NullReferenceException and not OperationCanceledException)
             {
                 HandleRemoteError(e, FeatureFlagError.UnknownError);
             }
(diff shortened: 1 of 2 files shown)


haacked (approved):
Nice work.

I recommend adding a test for user cancellation. Something like:

```csharp
[Fact]
public async Task PropagatesCancellationWhenUserCancelsRequest()
{
    var cts = new CancellationTokenSource();

    // Simulate: request starts, then user cancels
    container.FakeHttpMessageHandler.AddDecideResponseException(
        new TaskCanceledException("Cancelled", null, cts.Token));
    cts.Cancel();  // User's token is now cancelled

    // Should throw, NOT silently return false with "timeout" error
    await Assert.ThrowsAsync<TaskCanceledException>(() =>
        client.GetFeatureFlagAsync("flag", "id", null, cts.Token));
}
```

PR #181, opened 2026-04-20
fix: trim whitespace from API keys and host config

marandaneto:
This ports the whitespace-trimming fix from `posthog-go` to `posthog-dotnet`. It trims surrounding spaces and line breaks from the project API key and personal API key before using them so the SDK does not silently send invalid tokens or treat whitespace-only personal API keys as valid local-evaluation credentials.

- `dotnet test tests/UnitTests/UnitTests.csproj -f net8.0 --filter "FullyQualifiedName~LogsWarningWhenPersonalApiKeyIsBlankAfterTrimmingWhitespace|FullyQualifiedName~LogsErrorWhenProjectApiKeyIsBlankAfterTrimmingWhitespace|FullyQualifiedName~LogsWarningWhenPersonalApiKeyIsNull"`

## :pencil: Checklist

### If releasing new changes

- [x] Added the `release` label to the PR
- [x] Added exactly one version bump label: `bump-patch`, `bump-minor`, or `bump-major`

### DIFF
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -91,9 +91,7 @@ public sealed class PostHogClient : IPostHogClient
     static void NormalizeOptions(PostHogOptions options, ILogger<PostHogClient> logger)
     {
         options.ProjectApiKey = options.ProjectApiKey?.Trim();
-        options.PersonalApiKey = string.IsNullOrWhiteSpace(options.PersonalApiKey)
-            ? null
-            : options.PersonalApiKey.Trim();
+        options.PersonalApiKey = options.PersonalApiKey.NullIfEmpty();
 
         if (string.IsNullOrEmpty(options.ProjectApiKey))
         {
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -50,6 +50,9 @@ public sealed class PostHogClient : IPostHogClient
         _taskScheduler = taskScheduler ?? new TaskRunTaskScheduler();
         _timeProvider = timeProvider ?? TimeProvider.System;
         loggerFactory ??= NullLoggerFactory.Instance;
+        _logger = loggerFactory.CreateLogger<PostHogClient>();
+
+        NormalizeOptions(_options.Value, _logger);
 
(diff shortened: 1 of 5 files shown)


src/PostHog/PostHogClient.cs
```diff
@@ -82,10 +85,22 @@ public PostHogClient(
...
-        _logger = loggerFactory.CreateLogger<PostHogClient>();
         _logger.LogInfoClientCreated(options.Value.MaxBatchSize, options.Value.FlushInterval, options.Value.FlushAt);
     }
 
+    static void NormalizeOptions(PostHogOptions options, ILogger<PostHogClient> logger)
+    {
+        options.ProjectApiKey = options.ProjectApiKey?.Trim();
+        options.PersonalApiKey = string.IsNullOrWhiteSpace(options.PersonalApiKey)
+            ? null
+            : options.PersonalApiKey.Trim();
```
dustinbyrne, line 96: > Dereference of a possibly null reference.

Tooling fail, looks like the optional chain is needed

dustinbyrne: Weird that `dotnet` and CodeQL agree on this

dustinbyrne (approved):
approving to unblock, but the null check is needed

src/PostHog/PostHogClient.cs
haacked, line 96: I wonder if that method is missing `[NotNullWhen(true)]`. In either case, this is probably the easiest fix.

```suggestion
            : options.PersonalApiKey!.Trim();
```

haacked: Another way to go about this is to add a new extension method to src/PostHog/Library/StringExtensions.cs

```csharp
public static string? NullIfEmpty(this string? value)
{
    return string.IsNullOrEmpty(value) ? null : value.Trim();
}
```

Then you can do this everywhere: `options.PersonalApiKey.NullIfEmpty()`

haacked: Or call it `SafeTrim()`

marandaneto: i like the new extension idea since it can be reused

PR #183, opened 2026-04-22
Rename ProjectApiKey to ProjectToken with aliases

marandaneto:

The .NET SDK still exposed and documented `ProjectApiKey`, while other SDKs are moving toward `project token` terminology. This updates the public .NET surface to prefer `ProjectToken`, while keeping `ProjectApiKey` and legacy configuration keys working as backwards-compatible aliases.

It also adds clarifying comments around wire-level names like `api_key` and `token` so readers can see that those protocol fields still carry the project token.

This follow-up improves runtime logging for the deprecated alias by warning when `ProjectApiKey` is used, and makes missing-value messages clearer by using an explicit either/or message.

follow up from flutter https://github.com/PostHog/posthog-flutter/pull/374

```bash
dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --filter "FullyQualifiedName~PostHogClientTests.LogsWarningWhenProjectApiKeyIsUsed|FullyQualifiedName~PostHogClientTests.LogsErrorWhenProjectTokenIsMissing|FullyQualifiedName~RegistrationTests|FullyQualifiedName~RemoteConfigTests.IncludesProjectTokenInRemoteConfigUrl" --nologo
dotnet test tests/UnitTests.AspNetCore/UnitTests.AspNetCore.csproj --framework net8.0 --filter "FullyQualifiedName~RegistrationTests" --nologo
```

## :pencil: Checklist

### If releasing new changes

- [x] Added the `release` label to the PR
- [x] Added exactly one version bump label: `bump-patch`, `bump-minor`, or `bump-major`

### DIFF
--- a/samples/PostHog.Example.Console/Program.cs
+++ b/samples/PostHog.Example.Console/Program.cs
@@ -25,23 +25,24 @@ var envPaths = new[]
 DotEnv.Load(options: new DotEnvOptions(envFilePaths: envPaths, ignoreExceptions: true));
 
 // Get configuration from environment variables
-var projectApiKey = Environment.GetEnvironmentVariable("POSTHOG_PROJECT_API_KEY");
+var projectToken = Environment.GetEnvironmentVariable("POSTHOG_PROJECT_TOKEN")
+                   ?? Environment.GetEnvironmentVariable("POSTHOG_PROJECT_API_KEY");
 var personalApiKey = Environment.GetEnvironmentVariable("POSTHOG_PERSONAL_API_KEY");
 var endpoint = Environment.GetEnvironmentVariable("POSTHOG_HOST") ?? "https://us.i.posthog.com";
 
 // Check credentials
-if (string.IsNullOrEmpty(projectApiKey))
+if (string.IsNullOrEmpty(projectToken))
 {
-    Console.WriteLine("❌ Missing POSTHOG_PROJECT_API_KEY!");
+    Console.WriteLine("❌ Missing POSTHOG_PROJECT_TOKEN or POSTHOG_PROJECT_API_KEY!");
     Console.WriteLine("   Please set the environment variable or copy .env.example to .env");
     Console.WriteLine();
-    Console.Write("Enter your PostHog project API key (starts with phc_): ");
-    projectApiKey = Console.ReadLine()?.Trim();
+    Console.Write("Enter your PostHog project token (starts with phc_): ");
+    projectToken = Console.ReadLine()?.Trim();
 }
 
-if (string.IsNullOrEmpty(projectApiKey))
+if (string.IsNullOrEmpty(projectToken))
 {
-    Console.WriteLine("❌ Project API key is required. Exiting.");
+    Console.WriteLine("❌ Project token is required. Exiting.");
     return 1;
 }
 
@@ -55,7 +56,7 @@ if (endpointUri.Scheme != "https" && !endpointUri.IsLoopback)
 }
 
 Console.WriteLine("✅ PostHog credentials loaded successfully!");
-Console.WriteLine($"   Project API Key: ✅ Configured");
+Console.WriteLine($"   Project Token: ✅ Configured");
 Console.WriteLine($"   Personal API Key: {(string.IsNullOrEmpty(personalApiKey) ? "❌ Not set (local evaluation disabled)" : "✅ Configured")}");
 Console.WriteLine($"   Endpoint: {endpoint}");
 Console.WriteLine();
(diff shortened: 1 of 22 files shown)


marandaneto:
PersonalApiKey is correct already

turnipdabeets (approved):

samples/PostHog.Example.Console/Program.cs
```diff
@@ -25,23 +25,24 @@
...
 var personalApiKey = Environment.GetEnvironmentVariable("POSTHOG_PERSONAL_API_KEY");
 var endpoint = Environment.GetEnvironmentVariable("POSTHOG_HOST") ?? "https://us.i.posthog.com";
 
 // Check credentials
-if (string.IsNullOrEmpty(projectApiKey))
+if (string.IsNullOrEmpty(projectToken))
 {
-    Console.WriteLine("❌ Missing POSTHOG_PROJECT_API_KEY!");
+    Console.WriteLine("❌ Missing POSTHOG_PROJECT_TOKEN or POSTHOG_PROJECT_API_KEY!");
     Console.WriteLine("   Please set the environment variable or copy .env.example to .env");
```
rafaeelaudibert, line 37: Maybe we could update this to make it clear it's either/or? I know this says "or" already, but... people aren't good at reading.

sdk_compliance_adapter/Program.cs
```diff
@@ -154,6 +154,7 @@ record HealthResponse(
 );
 
 record InitRequest(
+    // The SDK compliance protocol still uses the wire name `api_key` for the PostHog project token.
```
rafaeelaudibert, line 157: This is a good thing actually @marandaneto. Can we update the compliance protocol first? It'll highlight which SDKs are incorrect and don't support passing `project_token` yet

src/PostHog/Api/PostHogApiClient.cs
```diff
@@ -69,7 +69,8 @@ public async Task<ApiResult> CaptureBatchAsync(
 
         var payload = new Dictionary<string, object>
         {
-            ["api_key"] = ProjectApiKey,
+            // PostHog's ingestion API still expects the project token under the wire name `api_key`.
+            ["api_key"] = ProjectToken,
```
rafaeelaudibert, line 73: Can we also raise this with ingestion and see if they can accept `token` as well?

rafaeelaudibert (approved):

sdk_compliance_adapter/Program.cs
marandaneto: SDK compliance protocol should match what the API is expecting, so this log is just to make it clearer to the reader, otherwise we'd have to project_token -> api_key in the SDK compliance

src/PostHog/Api/PostHogApiClient.cs
marandaneto: there's a new endpoint being developed which would replace this one so it's not worth a change for now
so once the new endpoint with new retry rules comes up, we can also add support for project_token
cc @eli-r-ph

marandaneto: i think `token` is already a valid property btw, capture checks for a lot of values to keep back compatibility, and i think `token` was one of them
but `token` != `project_token` :D

marandaneto: https://github.com/PostHog/posthog/blob/a35cc55914d063885d9f4a053e6372858bac9927/rust/capture/src/v0_request.rs#L29-L30

samples/PostHog.Example.Console/Program.cs
marandaneto: makes sense

samples/PostHog.Example.Console/Program.cs
```diff
@@ -25,23 +25,24 @@
...
+                   ?? Environment.GetEnvironmentVariable("POSTHOG_PROJECT_API_KEY");
 var personalApiKey = Environment.GetEnvironmentVariable("POSTHOG_PERSONAL_API_KEY");
 var endpoint = Environment.GetEnvironmentVariable("POSTHOG_HOST") ?? "https://us.i.posthog.com";
 
 // Check credentials
-if (string.IsNullOrEmpty(projectApiKey))
+if (string.IsNullOrEmpty(projectToken))
 {
-    Console.WriteLine("❌ Missing POSTHOG_PROJECT_API_KEY!");
+    Console.WriteLine("❌ Missing POSTHOG_PROJECT_TOKEN or POSTHOG_PROJECT_API_KEY!");
```
marandaneto:
```suggestion
    Console.WriteLine("❌ Missing either POSTHOG_PROJECT_TOKEN or POSTHOG_PROJECT_API_KEY!");
```

marandaneto:
runtime logging if they are using the older value https://github.com/PostHog/posthog-dotnet/pull/183/changes/c3c4aa28eb1590235366298d29a98ab18f3e4b93

PR #187, opened 2026-04-27
feat: add EvaluateFlagsAsync() API for single-call flag evaluation

dmarticus:
## Summary

- New `EvaluateFlagsAsync(distinctId, options)` returns a `FeatureFlagEvaluations` snapshot — one `/flags` call powers all flag branching for the rest of the request.
- The snapshot exposes `IsEnabled` / `GetFlag` / `GetFlagPayload` plus `OnlyAccessed()` and `Only(...)` filter helpers, with `$feature_flag_called` fired lazily on first access and deduped against the existing per-distinct-id cache.
- New `Capture(..., FeatureFlagEvaluations? flags, ...)` overload (and matching `CaptureException`) attaches `$feature/<key>` and `$active_feature_flags` from the snapshot — no second `/flags` request.
- Bug fix: locally-evaluated flags previously emitted `locally_evaluated=false` on `$feature_flag_called`. They now correctly carry `locally_evaluated=true`, `$feature_flag_reason="Evaluated locally"`, and a new `$feature_flag_definitions_loaded_at` timestamp.
- The single-flag API surface — `IsFeatureEnabledAsync`, `GetFeatureFlagAsync`, and the `bool sendFeatureFlags` overloads of `Capture`/`CaptureException` — is now `[Obsolete(error: false)]`. Users see a non-breaking migration warning on update.

RFC: [PostHog/requests-for-comments-internal#1020](https://github.com/PostHog/requests-for-comments-internal/pull/1020) · Reference implementations: [posthog-python#539](https://github.com/PostHog/posthog-python/pull/539), [posthog-js#3476](https://github.com/PostHog/posthog-js/pull/3476).

## Design decisions

- `GetFlag` returns the existing public `FeatureFlag?` (with implicit `bool`/`string` conversions) rather than mirroring Node's bare `bool|string|null` — keeps a single canonical .NET flag type and gives callers the payload too.
- `Capture(flags:)` is a separate overload, not a new optional parameter on the existing one — adding a defaulted parameter to a positional public signature is source-breaking.
- `AllFeatureFlagsOptions.FlagKeysToEvaluate` (already present) is reused as the request-body scoper; in-memory filtering uses the snapshot's own `Only(keys)` so the two concerns stay distinct.
- The snapshot takes a narrow internal `IFeatureFlagEvaluationsHost` (fire-deduped-event + log-warning), not the full client — keeps the snapshot easy to test in isolation.
- Filtered clones (`OnlyAccessed`/`Only`) get a fresh `ConcurrentDictionary` for accessed keys so child access doesn't back-propagate to the parent and the snapshot is safe to read across parallel branches.
- `OnlyAccessed()` with no prior access logs a warning and returns *all* flags as a fallback (rather than silently dropping exposure data); silenceable via a new `PostHogOptions.FeatureFlagsLogWarnings` knob.
- The new interface members are emitted as default interface methods on netstandard2.1+/net8 (delegating to the legacy overload) so external `IPostHogClient` implementers see an additive change. Netstandard2.0 keeps abstract members because the ns2.0 runtime can't host DIMs — .NET Framework 4.x consumers who hand-implement `IPostHogClient` (test fakes, custom decorators) will need to add the three new members on rebuild. The release notes should call this out.

## Phase 2 (separate PR)

- `PostHog.AspNetCore`'s `PostHogVariantFeatureManager` still calls the deprecated `GetFeatureFlagAsync` (per-flag) — migrate to `EvaluateFlagsAsync` + per-request snapshot.
- `PostHog.AI`'s `PostHogOpenAIHandler` still calls the deprecated `Capture(..., sendFeatureFlags: false, ...)`; switching to the snapshot overload also requires updating the NSubstitute test mocks.
- Migrate `Examples.cs` and the `samples/` projects to demonstrate the snapshot pattern in docs.

All UnitTests + UnitTests.AspNetCore + PostHog.AI.Tests pass; 18 new tests cover the snapshot, filter helpers, capture wiring, mixed local+remote merge, FlagMissing propagation, `CaptureException(flags:)`, and cross-path dedup.

> Originally opened as #186; force-pushing the rebase tripped the verified-signatures branch protection rule (an older unsigned auto-bot commit on main), so this is the same change on a fresh branch.

---


src/PostHog/IPostHogClient.cs
```diff
@@ -103,6 +103,27 @@ bool Capture(
...
+    /// event reflects the same flag values the caller branched on.
+    /// </summary>
+    /// <param name="distinctId">The identifier you use for the user.</param>
+    /// <param name="eventName">Human friendly name of the event. Recommended format [object] [verb] such as "Project created" or "User signed up".</param>
+    /// <param name="properties">Optional: The properties to send along with the event.</param>
+    /// <param name="groups">Optional: Context of what groups are related to this event, example: { ["company"] = "id:5" }. Can be used to analyze companies instead of users.</param>
+    /// <param name="flags">A snapshot of feature flag evaluations. When non-null, <c>$feature/&lt;key&gt;</c> and <c>$active_feature_flags</c> are attached from the snapshot — no <c>/flags</c> call is made.</param>
+    /// <param name="timestamp">Optional: Custom timestamp when the event occurred. If not provided, uses current time.</param>
+    /// <returns><c>true</c> if the event was successfully enqueued. Otherwise <c>false</c>.</returns>
+    bool Capture(
```
haacked, line 120: `blocking`: Adding three abstract members to a public interface is a source-and-binary break for any external implementer (custom decorators, hand-rolled fakes). `LangVersion 13.0` is set on this repo, so default interface methods are available. Default each new member to a sensible no-op and the change becomes additive.

```suggestion
    bool Capture(
        string distinctId,
        string eventName,
        Dictionary<string, object>? properties,
        GroupCollection? groups,
        FeatureFlagEvaluations? flags,
        DateTimeOffset? timestamp = null)
        => Capture(distinctId, eventName, properties, groups, sendFeatureFlags: false, timestamp);
```

Same shape on the `CaptureException` overload at line 155 (delegate to the `sendFeatureFlags: false` version) and on `EvaluateFlagsAsync` at line 231 (`throw new NotSupportedException(...)`). `PostHogClient`'s explicit implementations remain and shadow the defaults; in-box behavior is unchanged.

Directory.Build.props
```diff
@@ -1,6 +1,6 @@
 <Project>
     <PropertyGroup>
-        <Version>2.5.0</Version>
+        <Version>2.6.0</Version>
```
haacked: `blocking`: If the default-interface-methods change on `IPostHogClient` lands, 2.6.0 is fine. Without it, this should be 3.0.0. Adding non-default abstract members to a shipped public interface is a SemVer major break, and pinning to `[2.5.0,3.0.0)` shouldn't silently pick up a compile-breaking change.

```suggestion
        <Version>3.0.0</Version>
```

src/PostHog/Features/FeatureFlagEvaluations.cs
```diff
@@ -0,0 +1,231 @@
...
+            host,
+            distinctId,
+            new Dictionary<string, EvaluatedFlagRecord>(StringComparer.Ordinal),
+            requestId: null,
+            evaluatedAt: null,
+            flagDefinitionsLoadedAt: null,
+            groups: null,
+            errors: null);
+
+    EvaluatedFlagRecord? RecordAccess(string key)
```
haacked, line 200: `suggestion`: Every `IsEnabled`/`GetFlag` call hits `_host.TryCaptureFeatureFlagCalledEventIfNeeded`, which allocates a `List<string>`, a synthetic `FeatureFlag` (when `record` is null), a property `Dictionary`, and acquires the `MemoryCache` lock — even when the snapshot already saw this key. `_accessed.TryAdd` returns `false` on repeats; gate the host call on that. The `MemoryCache` still serves cross-snapshot dedup, so this is purely an in-snapshot fast path.

```suggestion
    EvaluatedFlagRecord? RecordAccess(string key)
    {
        var keyChecked = NotNull(key);
        var firstAccess = _accessed.TryAdd(keyChecked, 0);

        _records.TryGetValue(keyChecked, out var record);

        if (!firstAccess || string.IsNullOrEmpty(DistinctId))
        {
            // Already accessed in this snapshot, or empty-distinct-id safety fallback:
            // skip the per-call dedup cache hit and property allocation.
            return record;
        }

        _host.TryCaptureFeatureFlagCalledEventIfNeeded(
            distinctId: DistinctId,
            featureKey: keyChecked,
            record: record,
            groups: _groups,
            requestId: RequestId,
            evaluatedAt: EvaluatedAt,
            flagDefinitionsLoadedAt: FlagDefinitionsLoadedAt,
            errors: _errors);

        return record;
    }
```

For a request that calls `IsEnabled` 1000 times in a loop, this drops 1000 dictionary allocations and 1000 cache-lookup lock acquisitions to 1 of each.

src/PostHog/PostHogClient.cs
```diff
@@ -549,14 +612,245 @@ bool CaptureFeatureFlagCalledEvent(
...
+            var properties = BuildFeatureFlagCalledProperties(
+                featureKey,
+                flag,
+                requestId,
+                evaluatedAt,
+                snapshotErrors,
+                locallyEvaluated: record?.LocallyEvaluated ?? false,
+                flagDefinitionsLoadedAt: record?.LocallyEvaluated == true ? flagDefinitionsLoadedAt : null);
+
+            // For locally-evaluated flags from a snapshot we still want id/version/reason metadata
```
haacked: `suggestion`: For non-locally-evaluated flags, `BuildFeatureFlagCalledProperties` already writes `$feature_flag_id`/`$feature_flag_version`/`$feature_flag_reason` from the `FeatureFlagWithMetadata` branch. Then this block writes them again from `record.Id`/`record.Version`/`record.Reason` — same source (the record's `Flag` is the same `FeatureFlagWithMetadata` whose metadata seeded the record), so this is dead code. A reader has to trace both paths to convince themselves the values can't diverge.

If the intent is to handle a record that carries id/version without a `FeatureFlagWithMetadata` flag, make that intent explicit in a comment and have `BuildFeatureFlagCalledProperties` skip the metadata branch when called from this path. Otherwise drop lines 692–705.

src/PostHog/Features/EvaluatedFlagRecord.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    /// The string-form value used as the dedup-cache key for <c>$feature_flag_called</c>. Derived
+    /// from the implicit <see cref="FeatureFlag"/>-to-<see cref="string"/> conversion so the legacy
+    /// single-flag path and the snapshot path produce byte-identical cache keys.
+    /// </summary>
+    public required string CacheKeyValue { get; init; }
+
+    public int? Id { get; init; }
+    public int? Version { get; init; }
+    public string? Reason { get; init; }
```
haacked: `suggestion`: `Reason` semantics are split across three locations: this field holds the remote reason; `BuildFeatureFlagCalledProperties` hardcodes `"Evaluated locally"` for the local path; `EvaluationsHost.TryCaptureFeatureFlagCalledEventIfNeeded` then re-applies `record.Reason` only `if (!record.LocallyEvaluated)`. Three places to keep in sync.

Cleaner: store the resolved reason on the record at construction time. In `ToRecord`, set `reason = locallyEvaluated ? "Evaluated locally" : (flag as FeatureFlagWithMetadata)?.Reason`. Drop the local-vs-remote branch in `BuildFeatureFlagCalledProperties` and the `!record.LocallyEvaluated` guard in `EvaluationsHost`. The legacy single-flag path doesn't have a record so it would still hardcode the string inline — already the case today. The record becomes self-describing.

If you keep the current shape, at minimum document the asymmetry on the field:

```suggestion
    /// <summary>
    /// The evaluation reason from <see cref="FeatureFlagWithMetadata"/>. Intentionally <c>null</c>
    /// for locally-evaluated flags — <c>BuildFeatureFlagCalledProperties</c> hardcodes
    /// <c>"Evaluated locally"</c> for that path so the two reason sources cannot drift.
    /// </summary>
    public string? Reason { get; init; }
```

src/PostHog/Features/FeatureFlagEvaluations.cs
```diff
@@ -0,0 +1,231 @@
...
+    /// <summary>
+    /// The Unix-millisecond timestamp at which the local flag definitions used by the snapshot
+    /// were loaded, or <c>null</c> if no flag in the snapshot was locally evaluated.
+    /// </summary>
+    public long? FlagDefinitionsLoadedAt { get; }
+
+    /// <summary>
+    /// The set of flag keys present in this snapshot.
+    /// </summary>
+    public IReadOnlyCollection<string> Keys => (IReadOnlyCollection<string>)_records.Keys;
```
haacked: `suggestion`: `(IReadOnlyCollection<string>)_records.Keys` works today only because every construction path passes a `Dictionary<string, EvaluatedFlagRecord>`. A future maintainer who picks `ImmutableDictionary` or wraps `_records` in a custom view will get an `InvalidCastException` from a property they read as a plain getter.

Tighten the field to its actual type — it's internal anyway:

```suggestion
    readonly Dictionary<string, EvaluatedFlagRecord> _records;
```

The constructor signature would change to take `Dictionary<string, EvaluatedFlagRecord>`, which all callers already produce. Then `Keys => _records.Keys;` is a clean expression-bodied getter.

src/PostHog/PostHogClient.cs
```diff
@@ -298,6 +346,21 @@ static CapturedEvent AddFeatureFlagsToCapturedEvent(
         return capturedEvent;
     }
 
+    static CapturedEvent AddFeatureFlagsToCapturedEvent(
```
haacked, line 351: `suggestion`: This iterates `flags.Records` once for the per-flag properties, then again via LINQ (`Where().Select().ToArray()`) for `$active_feature_flags`. Combine into one pass and drop the LINQ enumerator allocations — runs per captured event, hot under load.

```suggestion
    static CapturedEvent AddFeatureFlagsToCapturedEvent(
        CapturedEvent capturedEvent,
        FeatureFlagEvaluations flags)
    {
        var active = new List<string>(flags.Records.Count);
        foreach (var (key, record) in flags.Records)
        {
            capturedEvent.Properties[$"$feature/{key}"] = record.Flag.ToResponseObject();
            if (record.Enabled)
            {
                active.Add(key);
            }
        }
        capturedEvent.Properties["$active_feature_flags"] = active.ToArray();
        return capturedEvent;
    }
```

src/PostHog/Features/FeatureFlagEvaluations.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                filtered[key] = record;
+            }
+            else
+            {
+                missing.Add(key);
+            }
+        }
+
+        if (missing.Count > 0)
```
haacked: `suggestion`: `string.Join(", ", missing)` runs even when `FeatureFlagsLogWarnings` is disabled — the host gates the call but the message is built first. Defer `missing` to lazy-init so the no-miss case is allocation-free regardless:

```suggestion
        if (missing is { Count: > 0 })
        {
            _host.LogFilterWarning(
                "FeatureFlagEvaluations.Only(...) requested keys that are not in the snapshot and will be dropped: " +
                string.Join(", ", missing));
        }
```

(Pair with declaring `missing` as `List<string>? missing = null;` and `(missing ??= new List<string>()).Add(key);` in the foreach above to skip the list allocation entirely on the no-miss path.)

src/PostHog/Features/FeatureFlagEvaluations.cs
```diff
@@ -0,0 +1,231 @@
...
+    public JsonDocument? GetFlagPayload(string key)
+        => _records.TryGetValue(NotNull(key), out var record) ? record.Flag.Payload : null;
+
+    /// <summary>
+    /// Returns a new snapshot containing only the flags that have been accessed via
+    /// <see cref="IsEnabled"/> or <see cref="GetFlag"/>. If no flags have been accessed yet,
+    /// logs a warning and returns a snapshot containing all flags so callers do not silently
+    /// drop exposure data.
+    /// </summary>
+    public FeatureFlagEvaluations OnlyAccessed()
```
haacked, line 121: `suggestion`: A method named `OnlyAccessed` that returns *all* flags when nothing has been accessed will trip callers. The XML doc says so, but that's a load-bearing doc for a load-bearing edge case. Two options worth considering: (a) return an empty snapshot — matches the literal name and surfaces "the caller forgot to access anything" as a real bug; or (b) keep the fallback but rename to `AccessedOrAll()` so the name encodes the policy. Your call.

If you keep the current shape, a `<remarks>` block on the existing doc would surface the fallback policy in IDE tooltips:

```suggestion
    /// <summary>
    /// Returns a new snapshot containing only the flags that have been accessed via
    /// <see cref="IsEnabled"/> or <see cref="GetFlag"/>.
    /// </summary>
    /// <remarks>
    /// <para>
    /// <b>Fallback behavior:</b> if no flags have been accessed yet, this method logs a warning and
    /// returns a snapshot containing <i>all</i> flags. This avoids silently dropping exposure data
    /// when callers wire up <c>OnlyAccessed()</c> before any branching logic runs. Set
    /// <see cref="PostHogOptions.FeatureFlagsLogWarnings"/> to <c>false</c> to suppress the warning.
    /// </para>
    /// </remarks>
    public FeatureFlagEvaluations OnlyAccessed()
```

src/PostHog/PostHogClient.cs
```diff
@@ -549,14 +612,245 @@ bool CaptureFeatureFlagCalledEvent(
...
+                }
+
+                if (flagsResult.QuotaLimited.Contains("feature_flags"))
+                {
+                    errors.Add(FeatureFlagError.QuotaLimited);
+                }
+
+                foreach (var (key, flag) in flagsResult.Flags)
+                {
+                    if (!records.ContainsKey(key))
```
haacked, line 806: `question`: `EvaluateFlagsAsync` does a local-wins merge here, but `GetAllFeatureFlagsAsync` (around line 877–893) does the opposite — it discards local results entirely on remote fallback. For flags that did evaluate locally but where the response also contains updated `id`/`version`/`reason`, the snapshot keeps the local (no metadata) version while `GetAllFeatureFlagsAsync` would surface the remote (with metadata) version.

Plausibly intentional (local-wins lets the snapshot tag those records `locally_evaluated=true` and add `$feature_flag_definitions_loaded_at`, which the legacy path can't), but neither code comment nor PR description calls out the divergence, and no test mixes local + remote for the same call. If intentional, a one-line comment would save future maintainers the trace:

```suggestion
                foreach (var (key, flag) in flagsResult.Flags)
                {
                    // Local-wins merge: keep the locally-evaluated record (which carries
                    // locally_evaluated=true and $feature_flag_definitions_loaded_at) and only
                    // fill in keys the local pass couldn't resolve. Differs from
                    // GetAllFeatureFlagsAsync, which discards local results entirely on fallback.
                    if (!records.ContainsKey(key))
                    {
                        records[key] = ToRecord(key, flag, locallyEvaluated: false);
                    }
                }
```

src/PostHog/PostHogClient.cs
```diff
@@ -549,14 +612,245 @@ bool CaptureFeatureFlagCalledEvent(
...
+                        flagDefinitionsLoadedAt = _featureFlagsLoader.FlagDefinitionsLoadedAt;
+                    }
+
+                    fallbackToRemote = needsRemote && options is not { OnlyEvaluateLocally: true };
+                }
+            }
+            catch (ApiException e) when (e.ErrorType is "quota_limited")
+            {
+                _logger.LogWarningQuotaExceeded(e);
+                return FeatureFlagEvaluations.Empty(_evaluationsHost, distinctId);
```
haacked: `question`: Local-pass quota_limited returns an empty snapshot here — discarding any flags evaluated up to this point and skipping the remote pass. But remote-pass quota_limited (around line 793–798) goes via `flagsResult.QuotaLimited.Contains("feature_flags")` → adds `FeatureFlagError.QuotaLimited` to the errors list and otherwise keeps the locally-evaluated records.

Local quota → empty snapshot, no error annotation. Remote quota → snapshot with whatever local produced, plus a `$feature_flag_error: "quota_limited"` annotation. The remote behavior is arguably the better one (preserves local work, surfaces the error). Is the discard on local-quota intentional?

src/PostHog/Features/IFeatureFlagEvaluationsHost.cs
```diff
@@ -0,0 +1,30 @@
...
+/// dedup cache and logger. The snapshot only needs these two operations, so it does not depend on
+/// the full <see cref="IPostHogClient"/> surface — keeping the snapshot simple and easy to test.
+/// </summary>
+internal interface IFeatureFlagEvaluationsHost
+{
+    /// <summary>
+    /// Fires a <c>$feature_flag_called</c> event for the given access, deduplicated against the
+    /// per-distinct-id cache that the legacy single-flag path also writes to.
+    /// </summary>
+    void TryCaptureFeatureFlagCalledEventIfNeeded(
```
haacked: `question`: Did you consider just taking an internal `PostHogClient` reference on the snapshot directly? Only `EvaluationsHost` implements this interface, the interface is internal, and no test substitutes a fake host. With a direct reference, `EvaluationsHost` and this file go away (~30 lines).

Defensible argument exists for the seam — keeps the snapshot honest about what state it depends on, future tests stay cheap to mock — but "future tests might want a mock" is the textbook YAGNI smell.

If you keep it, the method name itself is overstuffed: `Try`, `Capture`, `IfNeeded` are three uncertainty modifiers, and the method has no return value to make `Try` meaningful.

```suggestion
    void CaptureFeatureFlagCalled(
```

src/PostHog/IPostHogClient.cs
```diff
@@ -121,6 +142,24 @@ bool CaptureException(
...
+    /// Capture an exception as an event, attaching feature flag properties from a snapshot.
+    /// </summary>
+    /// <param name="exception">The exception to capture.</param>
+    /// <param name="distinctId">The identifier you use for the user.</param>
+    /// <param name="properties">Optional: The properties to send along with the event.</param>
+    /// <param name="groups">Optional: Context of what groups are related to this event.</param>
+    /// <param name="flags">A snapshot of feature flag evaluations. When non-null, <c>$feature/&lt;key&gt;</c> and <c>$active_feature_flags</c> are attached from the snapshot — no <c>/flags</c> call is made.</param>
+    /// <param name="timestamp">Optional: Custom timestamp when the event occurred.</param>
+    /// <returns><c>true</c> if the exception event was successfully enqueued. Otherwise <c>false</c>.</returns>
+    bool CaptureException(
```
haacked, line 163: `suggestion`: Consider landing `[Obsolete(error: false)]` on the legacy methods (`IsFeatureEnabledAsync`, `GetFeatureFlagAsync`, the `bool sendFeatureFlags` overload of `Capture`/`CaptureException`) in this PR rather than a follow-up. Non-breaking change, ~3 lines per method, lets users discover the migration path the moment they update the package.

```text
[Obsolete("Prefer EvaluateFlagsAsync(...) which fetches all flags in one call. See PR #187.", error: false)]
```

The PR description mentions this is deferred to a follow-up; nothing about adding `[Obsolete]` blocks the rest of the deprecation cleanup that follow-up would do (README, examples, etc.).

src/PostHog/IPostHogClient.cs
```diff
@@ -173,6 +212,27 @@ Task<IReadOnlyDictionary<string, FeatureFlag>> GetAllFeatureFlagsAsync(
...
+    /// <param name="distinctId">The identifier you use for the user.</param>
+    /// <param name="options">
+    /// Optional: Options used to control feature flag evaluation. <see cref="AllFeatureFlagsOptions.FlagKeysToEvaluate"/>
+    /// scopes the underlying <c>/flags</c> request body — distinct from
+    /// <see cref="FeatureFlagEvaluations.Only(System.Collections.Generic.IEnumerable{string})"/>, which
+    /// filters in memory.
+    /// </param>
+    /// <param name="cancellationToken">The cancellation token that can be used to cancel the operation.</param>
+    /// <returns>A snapshot of feature flag evaluations.</returns>
+    Task<FeatureFlagEvaluations> EvaluateFlagsAsync(
```
haacked, line 253: `suggestion`: Worth adding a `<remarks>` paragraph cross-referencing `Only(...)` so users see the distinction at the entry point too: `AllFeatureFlagsOptions.FlagKeysToEvaluate` scopes the underlying `/flags` request body. `FeatureFlagEvaluations.Only(keys)` filters an already-evaluated snapshot in memory. Two helpers, two costs — easy to confuse.

tests/UnitTests/Features/FeatureFlagEvaluationsTests.cs
```diff
@@ -0,0 +1,386 @@
...
+        // Snapshot path accesses the same flag — should hit the existing cache and NOT fire again.
+        var snapshot = await client.EvaluateFlagsAsync("user-1", options: null, CancellationToken.None);
+        snapshot.IsEnabled("flag-a");
+
+        await client.FlushAsync();
+        var body = batchHandler.GetReceivedRequestBody(indented: false);
+        var matches = System.Text.RegularExpressions.Regex.Matches(body, "\\$feature_flag_called");
+        Assert.Single(matches);
+    }
+}
```
haacked, line 470: `blocking`: Three coverage gaps for new code paths:

1. **Mixed local+remote merge.** `OnlyEvaluateLocallyDoesNotHitRemote` only exercises local. Nothing pins the merge in `EvaluateFlagsAsync` (PostHogClient.cs:800–806) where local fills some records and the remote pass fills the rest with `locallyEvaluated: false`. A regression that flips the `if (!records.ContainsKey(key))` guard wouldn't be caught.

2. **Error propagation.** `flagsResult.ErrorsWhileComputingFlags` adds `FeatureFlagError.ErrorsWhileComputingFlags` to the snapshot errors, and `EvaluationsHost` appends `FeatureFlagError.FlagMissing` when an unknown key is accessed. Neither path surfaces in any test, so the wiring through to `$feature_flag_error` on `$feature_flag_called` is unverified.

3. **`CaptureException(..., flags: snapshot)`.** Brand-new public overload, zero tests. A wiring mistake (e.g., accidentally passing `flags: null` in `CaptureExceptionCore`) would silently drop `$feature/<key>` props on exception events.

Skeleton for the merge test:

```csharp
[Fact]
public async Task MixedLocalAndRemoteEvaluationMergesRecordsAndTagsSourceCorrectly()
{
    var container = new TestContainer(personalApiKey: "fake-personal-api-key");
    container.FakeHttpMessageHandler.AddLocalEvaluationResponse(
        """{"flags": [
            {"id": 1, "key": "local-only", "active": true, "rollout_percentage": 100,
             "filters": {"groups": [{"properties": [], "rollout_percentage": 100}]}},
            {"id": 2, "key": "needs-remote",
             "filters": {"groups": [{"properties": [{"key": "country", "value": "US"}],
                                     "rollout_percentage": 100}]}}
        ]}""");
    container.FakeHttpMessageHandler.AddFlagsResponse(
        """{"featureFlags": {"needs-remote": "variant-x", "remote-only": true}}""");
    var batchHandler = container.FakeHttpMessageHandler.AddBatchResponse();

    var snapshot = await client.EvaluateFlagsAsync("user-1", options: null, CancellationToken.None);

    Assert.True(snapshot.IsEnabled("local-only"));
    Assert.Equal("variant-x", snapshot.GetFlag("needs-remote")?.VariantKey);
    Assert.True(snapshot.IsEnabled("remote-only"));

    await client.FlushAsync();
    using var doc = JsonDocument.Parse(batchHandler.GetReceivedRequestBody(indented: false));
    var byFlag = doc.RootElement.GetProperty("batch").EnumerateArray()
        .ToDictionary(
            e => e.GetProperty("properties").GetProperty("$feature_flag").GetString()!,
            e => e.GetProperty("properties").GetProperty("locally_evaluated").GetBoolean());
    Assert.True(byFlag["local-only"]);
    Assert.False(byFlag["needs-remote"]);
    Assert.False(byFlag["remote-only"]);
}
```

Skeleton for `FlagMissing` propagation:

```csharp
[Fact]
public async Task UnknownKeyAccessAppendsFlagMissingError()
{
    container.FakeHttpMessageHandler.AddFlagsResponse("""{"featureFlags": {"known": true}}""");
    var batchHandler = container.FakeHttpMessageHandler.AddBatchResponse();

    var snapshot = await client.EvaluateFlagsAsync("user-1", options: null, CancellationToken.None);
    snapshot.IsEnabled("does-not-exist");

    await client.FlushAsync();
    using var doc = JsonDocument.Parse(batchHandler.GetReceivedRequestBody(indented: false));
    var props = doc.RootElement.GetProperty("batch").EnumerateArray().Single().GetProperty("properties");
    Assert.Contains("FlagMissing",
        props.GetProperty("$feature_flag_error").GetString(), StringComparison.Ordinal);
}
```

Skeleton for `CaptureException`:

```csharp
[Fact]
public async Task CaptureExceptionAttachesFeatureFlagsFromSnapshot()
{
    container.FakeHttpMessageHandler.AddFlagsResponse(
        """{"featureFlags": {"flag-a": true, "flag-b": "variant-x"}}""");
    var batchHandler = container.FakeHttpMessageHandler.AddBatchResponse();

    var snapshot = await client.EvaluateFlagsAsync("user-1", options: null, CancellationToken.None);
    client.CaptureException(new InvalidOperationException("boom"), "user-1",
        properties: null, groups: null, flags: snapshot);

    await client.FlushAsync();
    using var doc = JsonDocument.Parse(batchHandler.GetReceivedRequestBody(indented: false));
    var exceptionEvent = doc.RootElement.GetProperty("batch").EnumerateArray()
        .Single(e => e.GetProperty("event").GetString() == "$exception");
    var props = exceptionEvent.GetProperty("properties");
    Assert.True(props.GetProperty("$feature/flag-a").GetBoolean());
    Assert.Equal("variant-x", props.GetProperty("$feature/flag-b").GetString());
}
```

haacked (commented):
Love seeming my friends write C# code! There's a couple minor blocking issues and a few other suggestions to make this nice idiomatic and non-breaking C#.

src/PostHog/IPostHogClient.cs
```diff
@@ -103,6 +103,31 @@ bool Capture(
...
+    /// <param name="timestamp">Optional: Custom timestamp when the event occurred. If not provided, uses current time.</param>
+    /// <returns><c>true</c> if the event was successfully enqueued. Otherwise <c>false</c>.</returns>
+    bool Capture(
+        string distinctId,
+        string eventName,
+        Dictionary<string, object>? properties,
+        GroupCollection? groups,
+        FeatureFlagEvaluations? flags,
+        DateTimeOffset? timestamp = null)
+#if !NETSTANDARD2_0
```
haacked, line 127: `question`: Worth flagging in the release notes that the DIM fix is partial. The `#if !NETSTANDARD2_0` guard wraps the default body but not the member itself, so on the netstandard2.0 build the new methods are still abstract. A .NET Framework consumer who hand-implements `IPostHogClient` (test fakes, decorators) will hit `CS0535` on rebuild. DIMs need IL features netstandard2.0 doesn't have, so the conditional is unavoidable here.

Given .NET Framework is the least supported audience, keeping 2.6.0 and calling this out in the release notes seems right. Something like:

> ⚠️ **Breaking for netstandard2.0 implementers.** Three new methods on `IPostHogClient` (`Capture(flags:)`, `CaptureException(flags:)`, `EvaluateFlagsAsync`) ship as abstract on the netstandard2.0 build because default interface methods aren't supported on that runtime. .NET Framework consumers who hand-implement `IPostHogClient` will need to add stubs (or upgrade to netstandard2.1+/net8.0 builds, where the members are default-implemented). netstandard2.1 and net8.0 consumers see this as an additive change.

src/PostHog/Features/FeatureFlagEvaluations.cs
```diff
@@ -0,0 +1,238 @@
...
+
+    /// <inheritdoc cref="Only(System.Collections.Generic.IEnumerable{string})"/>
+    public FeatureFlagEvaluations Only(params string[] keys)
+        => Only((IEnumerable<string>)NotNull(keys));
+
+    /// <summary>
+    /// The internal per-flag records. Used by <see cref="PostHogClient"/>'s capture path to attach
+    /// <c>$feature/&lt;key&gt;</c> properties.
+    /// </summary>
+    internal Dictionary<string, EvaluatedFlagRecord> Records => _records;
```
haacked: `suggestion`: Tightening the field to `Dictionary<>` was the right call for the `Keys` getter, but `Records` still hands the inner dictionary out by reference. The one call site (`PostHogClient.AddFeatureFlagsToCapturedEvent`) only iterates, so dropping to `IReadOnlyDictionary<>` here costs nothing and prevents a future `flags.Records.Add(...)` from silently corrupting the snapshot.

```suggestion
    internal IReadOnlyDictionary<string, EvaluatedFlagRecord> Records => _records;
```

src/PostHog/PostHogClient.cs
```diff
@@ -549,14 +616,218 @@ bool CaptureFeatureFlagCalledEvent(
...
+                    {
+                        flagDefinitionsLoadedAt = _featureFlagsLoader.FlagDefinitionsLoadedAt;
+                    }
+
+                    fallbackToRemote = needsRemote && options is not { OnlyEvaluateLocally: true };
+                }
+            }
+            catch (ApiException e) when (e.ErrorType is "quota_limited")
+            {
+                // Quota-limited at the local-evaluation endpoint: keep whatever local records we
```
haacked: `nit`: "keep whatever local records we computed before the failure" reads as if a partial set could be preserved, but `quota_limited` is only ever thrown by `_featureFlagsLoader.GetFeatureFlagsForLocalEvaluationAsync`, the first call in the `try`, before any evaluation runs. So `records` is always empty when this catch fires. Worth saying what the code actually does:

```suggestion
                // Quota-limited at /flags/definitions: no local records were produced (the failure
                // happens during the definitions fetch, before evaluation), so the snapshot ends
                // up empty. Still surface the error on $feature_flag_called and skip the remote
                // pass to avoid double-charging quota.
```

haacked (approved):
Nice! We'll make a .NET developer out of you!

Just some nits/suggestions. But nothing blocking.

PR #188, opened 2026-04-28
feat(flags): support mixed targeting in local evaluation

patricio-posthog:

Feature flags with "mixed targeting" (the `feature-flag-mixed-targeting` beta) allow different condition sets within a single flag to target different aggregation levels — some targeting users/persons, others targeting groups.

The SDK only reads `aggregation_group_type_index` at the flag level. For mixed flags this is `null`, so the SDK treats all conditions as person-targeted. Group conditions fail property matching and the SDK falls back to a server-side HTTP call.

This breaks customers using flags in environments that cannot make HTTP calls (e.g., background workers).

Ports [posthog-python#523](https://github.com/PostHog/posthog-python/pull/523) to posthog-dotnet.

Updated `MatchFeatureFlagProperties` in `LocalEvaluator` to resolve aggregation per condition when a condition sets its own `AggregationGroupTypeIndex` that differs from the flag level. Each condition uses the correct properties and bucketing value for its aggregation type. Backwards compatible with existing pure person and pure group flags.

Added a new `AggregationGroupTypeIndex` property on `FeatureFlagGroup` (internal record — not part of the public API surface, so non-breaking).

Added 5 new unit tests in `TheMixedTargetingEvaluation`: person match, group match, no match, only-group-no-groups, and rollout bucketing. All 767 existing UnitTests continue to pass on net8.0; xUnit CI run reports 805 passed, 0 failed, 2 skipped.

### Note on the SDK compliance report

The auto-posted `posthog-dotnet Compliance Report` comment shows `request_payload.request_with_person_properties_device_id` failing with a 404 from `http://sdk-adapter:8080/get_feature_flag`. **This failure pre-exists on `main` and is not related to this PR** — it's reproducible on the three most recent merged PRs (#182, #183, #184). The cause is that the .NET compliance adapter (`sdk_compliance_adapter/`) doesn't implement the `/get_feature_flag` endpoint yet (mirrors the same gap that posthog-go had). Worth tracking as a follow-up to bring the adapter to parity with the harness contract, but out of scope here.

## :pencil: Checklist

### If releasing new changes

- [ ] Added the `release` label to the PR
- [ ] Added exactly one version bump label: `bump-patch`, `bump-minor`, or `bump-major`

### DIFF
--- a/src/PostHog/Features/LocalEvaluator.cs
+++ b/src/PostHog/Features/LocalEvaluator.cs
@@ -279,8 +279,9 @@ internal sealed class LocalEvaluator
                 var effectiveProperties = properties;
                 var effectiveBucketingId = distinctId;
 
-                // Mixed-override path: condition-level aggregation differs from flag-level.
-                // This assumes flag-level aggregation is null for mixed flags.
+                // The condition explicitly sets its own aggregation, different from the
+                // flag level. Re-resolve properties and bucketing id from the condition's
+                // group so this condition evaluates against that group.
                 if (conditionAggregation != flagAggregation)
                 {
                     if (conditionAggregation.HasValue)
--- a/src/PostHog/Features/LocalEvaluator.cs
+++ b/src/PostHog/Features/LocalEvaluator.cs
@@ -282,20 +282,20 @@ internal sealed class LocalEvaluator
                 // The condition explicitly sets its own aggregation, different from the
                 // flag level. Re-resolve properties and bucketing id from the condition's
                 // group so this condition evaluates against that group.
+                // conditionAggregation is non-null here: if it were null and flagAggregation
+                // were also null they'd be equal, and the only way conditionAggregation can
+                // be null at all (after the ?? above) is that case.
                 if (conditionAggregation != flagAggregation)
                 {
-                    if (conditionAggregation.HasValue)
+                    if (!_groupTypeMapping.TryGetValue(conditionAggregation!.Value, out var groupType)
+                        || groups is null
+                        || !groups.TryGetGroup(groupType, out var group))
                     {
-                        if (!_groupTypeMapping.TryGetValue(conditionAggregation.Value, out var groupType)
-                            || groups is null
-                            || !groups.TryGetGroup(groupType, out var group))
-                        {
-                            // Skip this condition: group type unknown or not passed in.
-                            continue;
-                        }
-                        effectiveProperties = group.Properties;
-                        effectiveBucketingId = group.GroupKey;
+                        // Skip this condition: group type unknown or not passed in.
+                        continue;
                     }
+                    effectiveProperties = group.Properties;
+                    effectiveBucketingId = group.GroupKey;
                 }
 
                 // if any one condition resolves to True, we can short circuit and return
--- a/src/PostHog/Features/LocalEvaluator.cs
+++ b/src/PostHog/Features/LocalEvaluator.cs
@@ -262,6 +262,7 @@ internal sealed class LocalEvaluator
     {
         var filters = flag.Filters;
         var flagConditions = filters?.Groups ?? [];
+        var flagAggregation = filters?.AggregationGroupTypeIndex;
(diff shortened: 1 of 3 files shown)


tests/UnitTests/Features/LocalEvaluatorTests.cs
```diff
@@ -748,6 +749,211 @@ public void ThrowsInconclusiveMatchExceptionWhenUnknownOperator()
...
+                    Active = true,
+                    Filters = new FeatureFlagFilters
+                    {
+                        AggregationGroupTypeIndex = null,
+                        Groups = [
+                            new FeatureFlagGroup
+                            {
+                                AggregationGroupTypeIndex = 0,
+                                Properties = [],
+                                RolloutPercentage = 100
```
haacked: `suggestion`: At `RolloutPercentage = 100` and `Properties = []`, `IsConditionMatch` returns `true` at the rollout-100 short circuit before any hash runs. The flag has no `Multivariate` either, so `GetMatchingVariant` returns `null` and falls through. If someone reverted the `effectiveBucketingId` swap and left `distinctId` flowing through unchanged, this test would still pass and the regression would be invisible.

To actually verify group-key bucketing, use a partial rollout and pick a group key whose hash lands inside the bucket while the distinct id's hash lands outside (or vice versa). One fixture with two group keys, one in-bucket and one out, gives a clean cross-check:

```csharp
// Pre-computed against Hash("rollout-flag", …) so they straddle 50%.
const string flagKey = "rollout-flag";
const string distinctId = "user-1";
const string inBucketGroup  = "<group-key-whose-hash-is-below-0.5>";
const string outOfBucketGroup = "<group-key-whose-hash-is-above-0.5>";

var flags = new LocalEvaluationApiResult { /* RolloutPercentage = 50, AggregationGroupTypeIndex = 0, Properties = [] */ };
var evaluator = new LocalEvaluator(flags);

Assert.Equal(true,  evaluator.EvaluateFeatureFlag(flagKey, distinctId,
    groups: new GroupCollection { new Group("company", inBucketGroup) }));
Assert.Equal(false, evaluator.EvaluateFeatureFlag(flagKey, distinctId,
    groups: new GroupCollection { new Group("company", outOfBucketGroup) }));
```

The two group keys can be picked once with a small one-off helper that iterates candidates against the existing `Hash` function.

src/PostHog/Features/LocalEvaluator.cs
```diff
@@ -262,16 +262,44 @@ StringOrValue<bool> MatchFeatureFlagProperties(
...
+                // sets its own AggregationGroupTypeIndex (mixed targeting). When absent,
+                // fall back to the flag-level aggregation so existing pure person and
+                // pure group flags keep their original behavior.
+                var conditionAggregation = condition.AggregationGroupTypeIndex ?? flagAggregation;
+
+                var effectiveProperties = properties;
+                var effectiveBucketingId = distinctId;
+
+                // Mixed-override path: condition-level aggregation differs from flag-level.
+                // This assumes flag-level aggregation is null for mixed flags.
```
haacked: `suggestion`: The comment is narrower than the code. The branch is `if (conditionAggregation != flagAggregation)`, which also fires when a flag has a non-null `AggregationGroupTypeIndex` and a condition overrides to a different non-null index. The override correctly handles that case too (it loads the condition's group properties from the override path, while the caller already prepared the flag-level group's values). Reading the current comment as a load-bearing precondition is misleading; nothing in the code depends on flag-level aggregation being null.

```csharp
// Mixed targeting: the condition explicitly sets its own aggregation,
// different from the flag level. Re-resolve properties and bucketing id
// from the condition's group so this condition evaluates against that
// group.
```

haacked (approved):
Heh! Perhaps another C# developer convert! 😆 

Two suggestions inline: the new rollout test does not actually exercise group-key bucketing (rollout 100 short-circuits before hashing), and one comment about the override path is narrower than the code. Greptile already covered the unreachable `HasValue` guard.

PR #189, opened 2026-04-30
fix: Disable client when project token is missing

marandaneto:
## Summary
- Normalize ProjectToken and legacy ProjectApiKey by trimming and treating empty/whitespace as missing
- Add a `PostHogOptions.Disabled` user-configurable flag (default `false`) and compute the effective disabled state in `PostHogClient`
- No-op client calls when disabled instead of sending an empty api_key or throwing later
- Log a warning when public client methods are called while the SDK is disabled
- No-op public methods that require `personal_api_key` when it is not configured, including local-only feature flag evaluation and remote config payloads
- Add focused coverage for missing/empty/whitespace ProjectToken and legacy ProjectApiKey, disabled-call logging, and personal_api_key-protected methods

## Tests
- `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --filter "FullyQualifiedName~TheDisabledClient|FullyQualifiedName~ThePersonalApiKeyProtectedMethods|FullyQualifiedName~LogsWarningWhenPersonalApiKey|FullyQualifiedName~LogsErrorWhenProjectTokenIsMissing|FullyQualifiedName~LogsWarningWhenProjectApiKeyIsUsed"`
- `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0` (fails in existing CaptureException tests: `CaptureExceptionWithInvalidFilePathInStackFrame`, `CaptureExceptionCauseIOFailureEmptyContext`)

Note: requested context files `context.md` and `plan.md` were not present in this checkout.


marandaneto:
@PostHog/team-client-libraries and @haacked i think this is what SDKs should do
can you double check so i can port this fix across SDKs? (i have a bunch of draft PRs already but some of them didnt address all stuff i found during this one)

turnipdabeets (approved):

tests/UnitTests/PostHogClientTests.cs
```diff
@@ -1404,6 +1404,166 @@ private static void AssertContextEmpty(JsonElement frame)
...
+        Assert.False(captured);
+        Assert.False(capturedException);
+        Assert.False(isFeatureEnabled);
+        Assert.Null(featureFlag);
+        Assert.Null(remoteConfigPayload);
+        Assert.Empty(allFlags);
+        Assert.Empty(captureHandler.ReceivedRequests);
+        Assert.Empty(batchHandler.ReceivedRequests);
+        Assert.Empty(flagsHandler.ReceivedRequests);
+        Assert.Empty(localEvaluationHandler.ReceivedRequests);
```
haacked: `blocking`: Every prior `Capture` in this test returned `false` and short-circuited *before* enqueue, so the batch handler's queue is already empty when `FlushAsync` runs. `Assert.Empty(batchHandler.ReceivedRequests)` is satisfied even if `FlushAsync`'s own `IsDisabled` guard is removed entirely. The disabled-flush contract is asserted only via the warning-log substring (which has its own coverage gap, see other comment).

Add a separate test that exercises `FlushAsync` against a non-empty queue:

```csharp
[Fact]
public async Task FlushAsyncDoesNotSendQueuedEventsWhenDisabledAfterCapture()
{
    var container = new TestContainer(services =>
    {
        services.Configure<PostHogOptions>(options =>
        {
            options.ProjectToken = "fake-project-token";
        });
    });
    var batchHandler = container.FakeHttpMessageHandler.AddBatchResponse();

    Assert.True(client.Capture("distinct-id", "some-event"));
    container.GetRequiredService<IOptions<PostHogOptions>>().Value.Disabled = true;

    await client.FlushAsync();

    Assert.Empty(batchHandler.ReceivedRequests);
    AssertDisabledLog(container, nameof(PostHogClient.FlushAsync));
}
```

If mid-lifetime toggling is explicitly unsupported per the `init`-only suggestion above, then drop `await client.FlushAsync();` from this test — it isn't actually verifying anything.

tests/UnitTests/PostHogClientTests.cs
```diff
@@ -1404,6 +1404,166 @@ private static void AssertContextEmpty(JsonElement frame)
...
+        AssertDisabledLog(container, nameof(PostHogClient.IdentifyAsync));
+        AssertDisabledLog(container, nameof(PostHogClient.GroupIdentifyAsync));
+        AssertDisabledLog(container, nameof(PostHogClient.Capture));
+        AssertDisabledLog(container, nameof(PostHogClient.CaptureException));
+        AssertDisabledLog(container, nameof(PostHogClient.FlushAsync));
+        AssertDisabledLog(container, nameof(PostHogClient.IsFeatureEnabledAsync));
+        AssertDisabledLog(container, nameof(PostHogClient.GetFeatureFlagAsync));
+        AssertDisabledLog(container, nameof(PostHogClient.GetRemoteConfigPayloadAsync));
+        AssertDisabledLog(container, nameof(PostHogClient.GetAllFeatureFlagsAsync));
+        AssertDisabledLog(container, nameof(PostHogClient.LoadFeatureFlagsAsync));
```
haacked: `suggestion`: This theory exercises `null`, `""`, and `" \n\t "` for `ProjectToken`, but never asserts the constructor still emits `LogErrorProjectTokenRequired` (EventId 14) for the empty/whitespace inputs (the standalone `LogsErrorWhenProjectTokenIsMissing` test only covers `null`). A regression that moves the auto-disable into `Normalize()` but accidentally drops the error log for whitespace would slip through.

Add to the existing assertions:

```csharp
var errorLogs = container.FakeLoggerProvider.GetAllEvents(minimumLevel: LogLevel.Error);
Assert.Contains(errorLogs, log => log.Id.Id == 14);
```

src/PostHog/Config/PostHogOptions.cs
```diff
@@ -26,6 +27,15 @@ public string? ProjectToken
 
     internal bool HasLegacyProjectApiKey => _projectApiKey is not null;
 
+    internal void Normalize()
+    {
+        _projectToken = _projectToken.NullIfEmpty();
+        _projectApiKey = _projectApiKey.NullIfEmpty();
+        PersonalApiKey = PersonalApiKey.NullIfEmpty();
+        HostUrl = HostUrl.NormalizeHostUrl();
+        Disabled = Disabled || ProjectToken is null;
```
haacked: `suggestion`: This writes `Disabled = true` back onto the shared `PostHogOptions` instance that `IOptions<PostHogOptions>.Value` returns. Anyone else reading `IOptionsSnapshot<PostHogOptions>.Value.Disabled` (a health-check endpoint, a diagnostics page, an integration test, a second client constructed later) sees `Disabled = true` even though they never set it, with no way to tell whether that came from their config or from `Normalize`. The pre-PR mutations in `NormalizeOptions` only canonicalized values, so reading the option back gave callers the input they had passed in. This breaks that round-trip: `Disabled` no longer reflects what the user configured.

Pair with the `init`-only setter from the other comment, then compute the effective state on `PostHogClient` instead of writing back:

```suggestion
        _projectToken = _projectToken.NullIfEmpty();
        _projectApiKey = _projectApiKey.NullIfEmpty();
        PersonalApiKey = PersonalApiKey.NullIfEmpty();
        HostUrl = HostUrl.NormalizeHostUrl();
```

In `PostHogClient`'s ctor: `_isDisabled = _options.Value.Disabled || _options.Value.ProjectToken is null;`. `IsDisabled(method)` reads `_isDisabled`. That removes the leaked-state issue, the re-enable footgun, the `Normalize()` order-dependence, and the per-call `_options.Value` indirection, all in one move.

src/PostHog/PostHogClient.cs
```diff
@@ -95,16 +95,42 @@ static void NormalizeOptions(PostHogOptions options, ILogger<PostHogClient> logg
             logger.LogWarningProjectApiKeyDeprecated();
         }
 
-        options.ProjectToken = options.ProjectToken?.Trim();
-        options.PersonalApiKey = options.PersonalApiKey.NullIfEmpty();
-        options.HostUrl = options.HostUrl.NormalizeHostUrl();
+        options.Normalize();
 
-        if (string.IsNullOrEmpty(options.ProjectToken))
+        if (options.ProjectToken is null)
```
haacked: `suggestion`: After `options.Normalize()` runs on the line above, it has already computed `Disabled = Disabled || ProjectToken is null`. Re-checking `options.ProjectToken is null` here expresses the same intent twice. Future changes to "what counts as missing" must be made in both places.

```suggestion
        if (options.Disabled)
```

Caveat: this is misleading if a user explicitly sets `Disabled = true` with a valid token (the error log would falsely claim ProjectToken is required). Best paired with the derived-state refactor that splits "user-disabled" from "auto-disabled because token missing."

src/PostHog/PostHogClient.cs
```diff
@@ -95,16 +95,42 @@ static void NormalizeOptions(PostHogOptions options, ILogger<PostHogClient> logg
...
         {
             logger.LogErrorProjectTokenRequired();
         }
     }
 
+    bool IsDisabled(string methodName)
```
haacked: `suggestion`: `IsDisabled` reads as a pure query but emits a `Warning` log on every call. Same for `IsPersonalApiKeyMissing` below. Eleven call sites read `if (IsDisabled(nameof(X)))` like a benign check, so a future maintainer adding a twelfth caller (a debug path, a retry hook, telemetry) inadvertently emits a user-facing warning.

Either rename to make the side effect honest:

```suggestion
    bool CheckDisabledAndLog(string methodName)
```

…or split into a pure `bool Disabled => _options.Value.Disabled` property and a separate `LogDisabled(methodName)` call at each site.

src/PostHog/PostHogClient.cs
```diff
@@ -565,6 +653,16 @@ public async Task<IReadOnlyDictionary<string, FeatureFlag>> GetAllFeatureFlagsAs
         AllFeatureFlagsOptions? options,
         CancellationToken cancellationToken)
     {
+        if (IsDisabled(nameof(GetAllFeatureFlagsAsync)))
+        {
+            return new Dictionary<string, FeatureFlag>();
+        }
```
haacked, line 957: `suggestion`: `new Dictionary<string, FeatureFlag>()` allocates per disabled call. Same on line 664, and `new ApiResult(0)` on the disabled paths in `AliasAsync`, `IdentifyAsync`, both `GroupIdentifyAsync` overloads, and `CaptureException`. At 1k RPS on a misconfigured fleet of 1,000 instances that's ~48 KB/s of pure garbage per instance from the dictionary alone, on the same hot path that's already log-spamming (existing finding #5).

Hoist to static singletons:

```suggestion
            return EmptyFeatureFlags;
```

Where `EmptyFeatureFlags` is `private static readonly IReadOnlyDictionary<string, FeatureFlag> EmptyFeatureFlags = new Dictionary<string, FeatureFlag>(0);` (or `ImmutableDictionary<string, FeatureFlag>.Empty`) declared at class scope. Same pattern for `ApiResult`: `private static readonly Task<ApiResult> EmptyApiResultTask = Task.FromResult(new ApiResult(0));`.

src/PostHog/PostHogClient.cs
```diff
@@ -565,6 +653,16 @@ public async Task<IReadOnlyDictionary<string, FeatureFlag>> GetAllFeatureFlagsAs
         AllFeatureFlagsOptions? options,
         CancellationToken cancellationToken)
     {
+        if (IsDisabled(nameof(GetAllFeatureFlagsAsync)))
```
haacked: `suggestion`: This method is `async Task<…>` but the disabled path returns synchronously, so the C# compiler still allocates the async state machine and walks `ExecutionContext`. `GroupIdentifyAsync` (line 182) shows the pattern: `Task` return, no `async`, `Task.FromResult(...)` on the fast path, delegate to a private `async` core for the real work.

Same applies to `AliasAsync` (148), `IdentifyAsync` (163), `CaptureException` (277), `IsFeatureEnabledAsync` (369), `GetFeatureFlagAsync` (395), `GetRemoteConfigPayloadAsync` (545), `LoadFeatureFlagsAsync` (748), `GetLocalEvaluatorAsync` (798).

Pairs with the cached-result singletons above to make the disabled path truly zero-allocation.

tests/UnitTests/PostHogClientTests.cs
```diff
@@ -1404,6 +1404,166 @@ private static void AssertContextEmpty(JsonElement frame)
...
+
+        var options = ((IOptions<PostHogOptions>)((IServiceProvider)container).GetService(typeof(IOptions<PostHogOptions>))!).Value;
+        Assert.Null(options.ProjectToken);
```
haacked, line 1479: `blocking`: `AssertDisabledLog` uses `Contains` to match the method name, which leaves two gaps:

1. **Prefix collision**: `"Capture"` is a substring of `"CaptureException"`. If `IsDisabled` is removed from `Capture` but kept on `CaptureException`, `AssertDisabledLog(container, nameof(Capture))` still passes because the `CaptureException` emission contains `"Capture"`.
2. **Overload undercounting**: `NoOpsWhenProjectTokenIsMissingEmptyOrWhitespace` invokes `GroupIdentifyAsync` twice (two-arg and three-arg overloads) but asserts `AssertDisabledLog(container, nameof(GroupIdentifyAsync))` once, and a single emission satisfies it. Drop the guard from one overload and the suite still goes green.

Anchor on the EventId and the full `"{methodName} is a no-op"` template, and assert call counts:

```suggestion
    static void AssertDisabledLog(TestContainer container, string methodName, int expectedCount = 1)
    {
        var warningLogs = container.FakeLoggerProvider.GetAllEvents(minimumLevel: LogLevel.Warning);
        var matches = warningLogs.Count(log =>
            log.Id.Id == 20
            && log.Message?.Contains($"{methodName} is a no-op", StringComparison.Ordinal) == true);
        Assert.Equal(expectedCount, matches);
    }
```

Update the `GroupIdentifyAsync` call site to `expectedCount: 2`. Same shape for `AssertPersonalApiKeyMissingLog` below: anchor on `EventId == 21`.

tests/UnitTests/PostHogClientTests.cs
haacked, line 1463: `blocking`: Every prior `Capture` in this test returned `false` and short-circuited *before* enqueue, so the batch handler's queue is already empty when `FlushAsync` runs. `Assert.Empty(batchHandler.ReceivedRequests)` is satisfied even if `FlushAsync`'s own `IsDisabled` guard is removed entirely. The disabled-flush contract is asserted only via the warning-log substring (which has its own coverage gap, see other comment).

Add a separate test that exercises `FlushAsync` against a non-empty queue:

```csharp
[Fact]
public async Task FlushAsyncDoesNotSendQueuedEventsWhenDisabledAfterCapture()
{
    var container = new TestContainer(services =>
    {
        services.Configure<PostHogOptions>(options =>
        {
            options.ProjectToken = "fake-project-token";
        });
    });
    var batchHandler = container.FakeHttpMessageHandler.AddBatchResponse();

    Assert.True(client.Capture("distinct-id", "some-event"));
    container.GetRequiredService<IOptions<PostHogOptions>>().Value.Disabled = true;

    await client.FlushAsync();

    Assert.Empty(batchHandler.ReceivedRequests);
    AssertDisabledLog(container, nameof(PostHogClient.FlushAsync));
}
```

If mid-lifetime toggling is explicitly unsupported per the `init`-only suggestion above, then drop `await client.FlushAsync();` from this test. It isn't actually verifying anything.

tests/UnitTests/PostHogClientTests.cs
```diff
@@ -1404,6 +1404,166 @@ private static void AssertContextEmpty(JsonElement frame)
     }
 }
 
+public class TheDisabledClient
+{
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    [InlineData(" \n\t ")]
```
haacked, line 1413: `suggestion`: This theory pins the contract for ASCII whitespace only. `NullIfEmpty()` calls `string.Trim()` (no argument), which strips characters where `char.IsWhiteSpace` is true (including NBSP at U+00A0 and ideographic space at U+3000), but **not** zero-width space (U+200B). A future swap to `IsNullOrWhiteSpace` or a custom trim regex would silently shift behavior at these edges.

```suggestion
    [InlineData(" \n\t ")]
    [InlineData("\u00A0")]
```

The NBSP case documents that any `char.IsWhiteSpace`-true character normalizes to disabled. Worth a short comment too: "ZWSP (U+200B) is intentionally treated as a real token; it's not in `char.IsWhiteSpace`."

tests/UnitTests/PostHogClientTests.cs
haacked, line 1475: `suggestion`: This theory exercises `null`, `""`, and `" \n\t "` for `ProjectToken`, but never asserts the constructor still emits `LogErrorProjectTokenRequired` (EventId 14) for the empty/whitespace inputs (the standalone `LogsErrorWhenProjectTokenIsMissing` test only covers `null`). A regression that moves the auto-disable into `Normalize()` but accidentally drops the error log for whitespace would slip through.

Add to the existing assertions:

```csharp
var errorLogs = container.FakeLoggerProvider.GetAllEvents(minimumLevel: LogLevel.Error);
Assert.Contains(errorLogs, log => log.Id.Id == 14);
```

haacked (changes requested):
A few suggestions in-line.

src/PostHog/PostHogClient.cs
```diff
@@ -461,7 +545,12 @@ void HandleRemoteError(Exception ex, string errorType)
...
-        if (_options.Value.PersonalApiKey is null)
+        if (IsDisabled(nameof(GetRemoteConfigPayloadAsync)))
+        {
+            return null;
+        }
+
+        if (IsPersonalApiKeyMissing(nameof(GetRemoteConfigPayloadAsync)))
         {
             _logger.LogWarningPersonalApiKeyRequiredForRemoteConfigPayload();
             return null;
```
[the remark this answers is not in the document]
marandaneto: Fixed — removed the method-specific personal_api_key warnings so the unified no-op warning is emitted once.

src/PostHog/Config/PostHogOptions.cs
marandaneto: Fixed — Normalize() no longer writes back to Disabled. PostHogClient now computes the effective disabled state from the configured Disabled flag plus missing ProjectToken.

src/PostHog/PostHogClient.cs
marandaneto: Fixed — NormalizeOptions now returns the project-token-missing state and the constructor uses that to compute the effective disabled state.

src/PostHog/PostHogClient.cs
marandaneto: Fixed — renamed the side-effecting helpers to CheckDisabledAndLog and CheckPersonalApiKeyMissingAndLog.

src/PostHog/PostHogClient.cs
marandaneto: Fixed — added cached no-op ApiResult/Task and EmptyFeatureFlags singletons for no-op return paths.

src/PostHog/PostHogClient.cs
marandaneto: Partially addressed — cached the no-op return values. I kept the async method structure intact to keep this fix focused and avoid a large control-flow refactor.

tests/UnitTests/PostHogClientTests.cs
marandaneto: Fixed — assertions now match the exact EventId/message and verify expected counts, including both GroupIdentifyAsync overload calls.

tests/UnitTests/PostHogClientTests.cs
marandaneto: Addressed — added a focused FlushAsync disabled test and tightened the disabled-log assertion. Effective disabled state is computed at init, so mid-lifetime toggling is not supported here.

tests/UnitTests/PostHogClientTests.cs
marandaneto: Fixed — added NBSP coverage and a comment documenting the Trim/ZWSP behavior.

tests/UnitTests/PostHogClientTests.cs
marandaneto: Fixed — the missing-token tests now assert the ProjectToken-required error EventId.

marandaneto:
> A few suggestions in-line.

Thanks and done.
can you double check?

dustinbyrne (approved):

src/PostHog/PostHogClient.cs
```diff
@@ -90,21 +96,49 @@ public PostHogClient(
...
-        options.ProjectToken = options.ProjectToken?.Trim();
-        options.PersonalApiKey = options.PersonalApiKey.NullIfEmpty();
-        options.HostUrl = options.HostUrl.NormalizeHostUrl();
+        options.Normalize();
 
-        if (string.IsNullOrEmpty(options.ProjectToken))
+        var projectTokenMissing = options.ProjectToken is null;
+        if (projectTokenMissing)
         {
             logger.LogErrorProjectTokenRequired();
```
haacked, line 111: `suggestion`: When a user explicitly sets `Disabled = true` and omits `ProjectToken` (a reasonable combo for "turn this off in tests" or "feature-flag the SDK off in this environment"), this still emits the `ProjectToken is required` error. The user did the correct thing and the SDK still complains.

```suggestion
        var projectTokenMissing = options.ProjectToken is null;
        if (projectTokenMissing && !options.Disabled)
        {
            logger.LogErrorProjectTokenRequired();
        }
        return projectTokenMissing;
```

The per-call `LogWarningClientDisabled` from `CheckDisabledAndLog` still surfaces "this client is disabled" at every call site, so the operator isn't left in the dark — they just don't get a misleading error claiming the token is the problem.

tests/UnitTests/PostHogClientTests.cs
```diff
@@ -1405,6 +1405,210 @@ private static void AssertContextEmpty(JsonElement frame)
...
+
+        Assert.Equal(0, identifyResult.Status);
+        Assert.False(captured);
+        Assert.False(isFeatureEnabled);
+        Assert.Empty(allFlags);
+        Assert.Empty(captureHandler.ReceivedRequests);
+        Assert.Empty(batchHandler.ReceivedRequests);
+        Assert.Empty(flagsHandler.ReceivedRequests);
+        Assert.Empty(localEvaluationHandler.ReceivedRequests);
+        AssertDisabledLog(container, nameof(PostHogClient.Capture));
```
haacked, line 1570: `suggestion`: This test invokes `IdentifyAsync`, `Capture`, `FlushAsync`, `IsFeatureEnabledAsync`, `GetAllFeatureFlagsAsync`, and `LoadFeatureFlagsAsync`, but only asserts the disabled log for `Capture`. A regression that drops `CheckDisabledAndLog` from any of the other five would slip through this test (the empty-handler assertions are satisfied because nothing was enqueued, same vacuity pattern as the original `FlushAsync` test).

The sibling `NoOpsWhenProjectTokenIsMissingEmptyOrWhitespace` covers all twelve methods, so this isn't a coverage hole in the suite — but this test, in isolation, isn't pulling its weight. Either mirror the full assertion set:

```suggestion
        AssertDisabledLog(container, nameof(PostHogClient.IdentifyAsync));
        AssertDisabledLog(container, nameof(PostHogClient.Capture));
        AssertDisabledLog(container, nameof(PostHogClient.FlushAsync));
        AssertDisabledLog(container, nameof(PostHogClient.IsFeatureEnabledAsync));
        AssertDisabledLog(container, nameof(PostHogClient.GetAllFeatureFlagsAsync));
        AssertDisabledLog(container, nameof(PostHogClient.LoadFeatureFlagsAsync));
```

…or shrink the test to just the legacy-key normalization contract (one method invocation, plus the `Assert.Null(options.ProjectToken)` round-trip) and leave the per-method coverage to the other test.

src/PostHog/Config/PostHogOptions.cs
```diff
@@ -49,6 +58,11 @@ public string? ProjectApiKey
     /// </remarks>
     public string? PersonalApiKey { get; set; }
 
+    /// <summary>
+    /// Whether this client is disabled and should no-op instead of sending data to PostHog. (Default: false)
+    /// </summary>
+    public bool Disabled { get; set; }
```
haacked, line 68: `suggestion`: `_isDisabled` is captured in the `PostHogClient` ctor, so a later `options.Disabled = true` (e.g. via `IOptionsSnapshot<PostHogOptions>.Value.Disabled = true` in a test, or a config-reload code path) is silently a no-op. The setter staying public invites that mistake.

Two cheap ways to address this:

1. Make `Disabled` init only.

```suggestion
    /// <summary>
    /// Whether this client is disabled and should no-op instead of sending data to PostHog. (Default: false)
    /// </summary>
    /// <remarks>
    /// Read once during <see cref="PostHogClient"/> construction. Mutating this after the client is
    /// activated has no effect. Re-create the client to change the disabled state.
    /// </remarks>
    public bool Disabled { get; init; }
```

…or keep `set` and add the same `<remarks>` block. Either way, callers learn the lifecycle from the type rather than from a PR comment thread.

I prefer the first as it enforces the behavior through the type system.

haacked (approved):
Some nits, but nothing major.

src/PostHog/PostHogClient.cs
marandaneto: Good catch! Fixed in c83eaad — missing project token now only logs an error when the client is not explicitly disabled, and I added coverage for disabled + missing/blank token still no-oping with the disabled warning.

tests/UnitTests/PostHogClientTests.cs
marandaneto: Fixed in f7d07f5 — mirrored the disabled-log assertions for every method this legacy-key test invokes, so each call now proves it goes through CheckDisabledAndLog.

src/PostHog/Config/PostHogOptions.cs
marandaneto: Fixed in 0d3756b by adding the lifecycle remarks while keeping the public setter for Configure<PostHogOptions> / options-pattern compatibility. Ideally these options would be immutable after creation, but that is a broader options-model change than this PR should take on.\n\nAlso filed https://github.com/PostHog/posthog-dotnet/issues/197 to track adding runtime OptOut / OptIn / IsOptOut support like the other SDKs, which should be the right API for changing this behavior after client construction.

PR #196, opened 2026-05-05
Allow PostHog packages to release individually

marandaneto:
A changeset that only bumped `PostHog` caused `PostHog.AspNetCore` and `PostHog.AI` to be versioned and released too, because their Changesets metadata used `workspace:*` for the internal `PostHog` dependency.

Switching those internal dependency ranges to `workspace:^` keeps compatible `PostHog` patch/minor releases from forcing dependency-only bumps of the dependent packages, so each package can be released individually unless it has its own changeset or the dependency update falls outside the compatible range.

- `pnpm install --frozen-lockfile`
- Temporarily added a test changeset for only `PostHog`, ran `pnpm changeset version`, and verified only `src/PostHog/package.json` / `src/PostHog/CHANGELOG.md` changed for the release; `PostHog.AspNetCore` and `PostHog.AI` were not version-bumped. Reverted the temporary generated files before committing.
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -12,7 +12,7 @@ This repo publishes three NuGet packages independently:
 | `PostHog.AspNetCore` | `PostHog.AspNetCore` | `src/PostHog.AspNetCore/package.json` | `src/PostHog.AspNetCore/PostHog.AspNetCore.csproj` |
 | `PostHog.AI` | `PostHog.AI` | `src/PostHog.AI/package.json` | `src/PostHog.AI/PostHog.AI.csproj` |
 
-`PostHog.AspNetCore` and `PostHog.AI` depend on `PostHog`. If a changeset bumps `PostHog`, Changesets will also patch-bump the dependent packages so their NuGet dependencies point at the new `PostHog` version.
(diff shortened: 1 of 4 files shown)


marandaneto:
@PostHog/team-client-libraries on the other hand this is quite nice, so not sure if we should keep as is and change the other monorepos to do this or if its better to be controlled, wdyt?
found by https://github.com/PostHog/posthog-dotnet/commit/92dd0e90162527495d41ce76a2eeaff9fcfd7164

turnipdabeets (approved):

ioannisj (approved):
> @PostHog/team-client-libraries on the other hand this is quite nice, so not sure if we should keep as is and change the other monorepos to do this or if its better to be controlled, wdyt? found by [92dd0e9](https://github.com/PostHog/posthog-dotnet/commit/92dd0e90162527495d41ce76a2eeaff9fcfd7164)

For me it makes a lot more sense for this to be controlled. So a base package update should not always release dependent packages. Otherwise the versions are too chained together, and a potential bug in a base package version will propagate to all packages immediately. 

We'll also need to update the note in [RELEASING.md](https://github.com/PostHog/posthog-dotnet/blame/92dd0e90162527495d41ce76a2eeaff9fcfd7164/RELEASING.md#L75) as well

PR #198, opened 2026-05-06
feat: Add server-side request context support

dustinbyrne:

Add ASP.NET Core request-scoped PostHog context so server-side captures and exception events during a request can inherit useful request metadata and, when enabled, PostHog tracing headers (`X-PostHog-Distinct-Id`, `X-PostHog-Session-Id`).

This aligns .NET behavior with the server-side request context guidelines while keeping the context implementation internal to avoid expanding the public .NET API surface.

Middleware behavior:

- `UsePostHogRequestContext` now always wraps ASP.NET Core requests and applies request metadata (`$current_url`, `$request_method`, `$request_path`, `$user_agent`, `$ip`) to captures made during the request.
- PostHog tracing header identity/session usage is controlled separately via `options.UseTracingHeaders`.
- Exception capture only adds exception-specific metadata; request context identity/session/properties are applied by the core capture path. ASP.NET Core `HttpContext.User` is not treated as a PostHog distinct ID automatically.

- `dotnet build PostHog.sln --no-restore`
- `dotnet test tests/UnitTests.AspNetCore/UnitTests.AspNetCore.csproj --no-restore`
- `dotnet test tests/UnitTests/UnitTests.csproj --no-restore --framework net8.0 --filter "FullyQualifiedName~PostHogContextTests|FullyQualifiedName~FeatureFlagEvaluationsTests"`

Note: full `UnitTests` net8 currently has two unrelated exception stack-frame assertion failures in this environment.

## :pencil: Checklist

### If releasing new changes


src/PostHog/Capture/CaptureExtensions.cs
```diff
@@ -9,6 +9,48 @@ namespace PostHog;
...
+    /// Captures an event using the current request context distinct ID, or as a personless event if none is set.
+    /// </summary>
+    /// <param name="client">The <see cref="IPostHogClient"/>.</param>
+    /// <param name="eventName">Human friendly name of the event.</param>
+    /// <param name="properties">Optional: The properties to send along with the event.</param>
+    /// <returns><c>true</c> if the event was successfully enqueued. Otherwise <c>false</c>.</returns>
+    public static bool Capture(
+        this IPostHogClient client,
+        string eventName,
+        Dictionary<string, object>? properties)
```
dustinbyrne: Worth a heads up, and I'm kind of torn on this here.

This overload can create ambiguity in the case of `Capture("id", null)`. There's similar ambiguity for `CaptureException(exception, null)` and `EvaluateFlagsAsync(null)`.

So from a source compatibility standpoint, these overloads aren't strictly backwards compatible, but I feel like it's enough of an edge case (and these overloads are useful enough) that it's worth keeping them. Open to thoughts here.

dustinbyrne: I was able to confine this to ASP.NET here 8b2c333bd5a6ca8f7da292c06e0b4d07f80fd299

It doesn't fully eliminate the ambiguity, but it does narrow the scope

src/PostHog.AspNetCore/Tracing/PostHogRequestContextMiddlewareExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Adds request-scoped PostHog context extraction for PostHog tracing headers and request metadata.
+    /// </summary>
+    /// <param name="app">The application builder.</param>
+    /// <param name="configure">Optional configuration for tracing headers and exception capture.</param>
+    /// <returns>The passed in <see cref="IApplicationBuilder" />.</returns>
+    public static IApplicationBuilder UsePostHogRequestContext(
+        this IApplicationBuilder app,
+        Action<PostHogRequestContextOptions>? configure = null)
+    {
+        ArgumentNullException.ThrowIfNull(app);
```
dustinbyrne, line 25: We typically avoid throwing, but this one seems unavoidable. If `app` is null, there's no easy way to recover.

Considering this would be called on initialization, avoiding an exception here could mask some serious issues.

src/PostHog.AspNetCore/Tracing/PostHogRequestPropertyNames.cs
```diff
@@ -0,0 +1,12 @@
+namespace PostHog;
+
+internal static class PostHogRequestPropertyNames
+{
+    public const string SessionId = "$session_id";
```
haacked: `suggestion`: `$session_id` is now defined twice: here as `PostHogRequestPropertyNames.SessionId` and in `PostHogProperties.SessionId` at `src/PostHog/Api/PostHogProperties.cs:10`. This makes me worry about future drift.

`PostHogProperties` is `public static class` and already reachable from `PostHog.AspNetCore`, so it makes sense to drop the duplicate here:

```suggestion
internal static class PostHogRequestPropertyNames
{
    public const string CurrentUrl = "$current_url";
    public const string RequestMethod = "$request_method";
    public const string RequestPath = "$request_path";
    public const string UserAgent = "$user_agent";
    public const string Ip = "$ip";
    public const string ResponseStatusCode = "$response_status_code";
}
```

…and at `PostHogTracingHeaders.cs:57`, replace `PostHogRequestPropertyNames.SessionId` with `PostHogProperties.SessionId`.

src/PostHog/PostHogClient.cs
```diff
@@ -486,6 +503,13 @@ public async Task<bool> IsFeatureEnabledAsync(
             return null;
         }
 
+        var resolvedDistinctId = PostHogContextHelper.ResolveDistinctId(distinctId);
```
haacked, line 506: `suggestion`: `IsFeatureEnabledAsync`/`GetFeatureFlagAsync` here and `GetAllFeatureFlagsAsync` at line 990 short-circuit on `resolvedDistinctId is null`, but `EvaluateFlagsAsync` at line 858 short-circuits on `resolvedDistinctId.Length == 0`.

`PostHogContextHelper.ResolveDistinctId("")` returns the empty string (it only falls through to context when the input is `null`), so a caller passing `""` skips the warning and proceeds to `/flags` with an empty distinct id from the first two methods, but gets the empty short-circuit and the warning from the third. Same package, same intent, three flag entrypoints, two different rules. Tighten `ResolveDistinctId` to use `IsNullOrEmpty` for the fallback decision so the three sites behave identically:

```csharp
// PostHogContext.cs
internal static string? ResolveDistinctId(string? preferredDistinctId = null)
    => !string.IsNullOrEmpty(preferredDistinctId) ? preferredDistinctId : PostHogContext.Current?.DistinctId;
```

Then `EvaluateFlagsAsync` can drop its separate `Length == 0` branch and rely on the null check the other two callers use.

src/PostHog/PostHogContext.cs
```diff
@@ -0,0 +1,169 @@
...
+    static readonly AsyncLocal<PostHogContext?> CurrentContext = new();
+
+    PostHogContext(
+        string? distinctId,
+        string? sessionId,
+        IReadOnlyDictionary<string, object>? properties)
+    {
+        DistinctId = string.IsNullOrEmpty(distinctId) ? null : distinctId;
+        SessionId = string.IsNullOrEmpty(sessionId) ? null : sessionId;
+        Properties = properties?.ToDictionary(pair => pair.Key, pair => pair.Value) ?? new Dictionary<string, object>(0);
```
haacked: `suggestion`: The `PostHogContext` constructor calls `properties?.ToDictionary(...)` to clone the incoming dictionary, but the only caller (`BeginScope` a few lines below) already builds a fresh `mergedProperties` dict it owns and hands it in. The defensive copy is pure waste on the request hot path: every `BeginScope` call copies the property dict twice in succession. Drop the ctor copy:

```suggestion
        Properties = properties ?? (IReadOnlyDictionary<string, object>)new Dictionary<string, object>(0);
```

src/PostHog/PostHogContext.cs
```diff
@@ -0,0 +1,169 @@
...
+    static Dictionary<string, object>? ApplyContextProperties(
+        Dictionary<string, object>? properties,
+        PostHogContext? context)
+    {
+        if (context is null)
+        {
+            return properties;
+        }
+
+        var mergedProperties = context.Properties.ToDictionary(pair => pair.Key, pair => pair.Value);
```
haacked: `suggestion`: `ApplyContextProperties` calls `context.Properties.ToDictionary(...)` for every capture, even when the caller passed no `properties` and no session merge is needed. For an ASP.NET request that captures 5 events, that's 5 default-capacity dictionary allocations of ~6 entries that immediately resize. Build the dict with the right capacity up front and skip the session-id check when there isn't one to add:

```suggestion
        var contextProps = context.Properties;
        var needsSession = !string.IsNullOrEmpty(context.SessionId)
            && !contextProps.ContainsKey(PostHogProperties.SessionId);

        var capacity = contextProps.Count + (needsSession ? 1 : 0) + (properties?.Count ?? 0);
        var mergedProperties = new Dictionary<string, object>(capacity);
        foreach (var (key, value) in contextProps)
        {
            mergedProperties[key] = value;
        }
        if (needsSession)
        {
            mergedProperties[PostHogProperties.SessionId] = context.SessionId!;
        }

        if (properties is not null)
        {
            foreach (var (key, value) in properties)
            {
                mergedProperties[key] = value;
            }
        }

        return mergedProperties;
```

The presized constructor avoiding rehash is the main win; the LINQ removal is secondary.

src/PostHog.AspNetCore/Tracing/PostHogTracingHeaders.cs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    internal static string? SanitizeValue(string? value)
+    {
+        if (string.IsNullOrWhiteSpace(value))
+        {
+            return null;
+        }
+
+        var sanitized = new string(value.Where(character => !char.IsControl(character)).ToArray()).Trim();
```
haacked, line 88: `suggestion`: This is 4 allocations when 1 will do. The `Where(...).ToArray()` chain builds an `IEnumerable<char>` enumerator, then a `char[]`, then a `string`, and `.Trim()` produces a fourth string. A single-pass loop into a `Span<char>` buffer collapses that to one allocation (the final string). Filter controls before truncation, not after, so the result for inputs with leading control characters stays the same as today.

```suggestion
    internal static string? SanitizeValue(string? value)
    {
        if (string.IsNullOrWhiteSpace(value))
        {
            return null;
        }

        Span<char> buffer = stackalloc char[MaxHeaderValueLength];
        var written = 0;
        foreach (var c in value!)
        {
            if (char.IsControl(c))
            {
                continue;
            }
            if (written == MaxHeaderValueLength)
            {
                break;
            }
            buffer[written++] = c;
        }

        var trimmed = buffer[..written].Trim();
        return trimmed.IsEmpty ? null : new string(trimmed);
    }
```

Worth a benchmark before merging; the gain only matters if header sanitization shows up in a profile.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+            groups: null,
+            flags: null);
+    }
+
+    /// <summary>
+    /// Evaluates all feature flags using the current request context distinct ID and returns a <see cref="FeatureFlagEvaluations"/> snapshot.
+    /// </summary>
+    /// <param name="client">The <see cref="IPostHogClient"/>.</param>
+    /// <returns>A snapshot of feature flag evaluations.</returns>
+    public static Task<FeatureFlagEvaluations> EvaluateFlagsAsync(this IPostHogClient client)
```
haacked, line 85: `suggestion`: The parameterless `EvaluateFlagsAsync(this IPostHogClient client)` here (and the `(client, options)` overload below) silently use `PostHogContextHelper.ResolveDistinctId() ?? string.Empty`, which on the ASP.NET Core path is whatever arrived in `X-POSTHOG-DISTINCT-ID`. The capture extensions above (`Capture`, `CaptureException`) do the same via `ResolveCaptureContext(distinctId: null, ...)`. With `UseTracingHeaders` defaulting to `true`, every parameterless call in this file is keyed on a client-controlled header.

Two concrete harms:

1. **Capture attribution**: anyone hitting the app can send `X-POSTHOG-DISTINCT-ID: ceo@company.com` and have server-side `Capture(...)` calls land on that person profile. Cohorts, funnels, and replays keyed on that distinct ID get polluted with attacker activity.

2. **Flag-eval authorization**: a developer reading IntelliSense writes `if ((await posthog.EvaluateFlagsAsync()).IsEnabled("admin-feature")) { ... }` thinking it checks the authenticated user. It checks whatever the client sent. Flag values aren't secret (anyone with the project's public key can call `/flags` directly), but the attacker now picks which user's flags decide the developer's `if`. That's an authorization bypass in the developer's app, not disclosure in PostHog.

A doc-only remarks block doesn't help; readers who skip XML docs still walk into both surfaces. Two structural options:

1. **Remove the parameterless `EvaluateFlagsAsync` overloads.** Force callers to pass a distinct ID (`posthog.EvaluateFlagsAsync(User.Identity.Name)`) so the compiler makes them pick a source. Capture extensions can stay since attribution drift is the lesser harm and the convenience matters.

2. **Default `UseTracingHeaders` to `false`.** Apps that want JS/server correlation opt in. The XML doc already calls these "client-supplied", so flipping the default matches the wording. This fixes both surfaces in one change.

I'd take option 1 first since the flag-eval footgun is the sharper edge, and add option 2 if you're willing. At minimum, the README should call the spoofing surface out so an operator turning `UseTracingHeaders` on knows what they're accepting.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextMiddleware.cs
```diff
@@ -0,0 +1,107 @@
...
+/// <summary>
+/// ASP.NET Core middleware that applies PostHog request context to the current async request flow.
+/// </summary>
+internal sealed class PostHogRequestContextMiddleware(
+    RequestDelegate next,
+    IPostHogClient postHog,
+    PostHogRequestContextOptions options)
+{
+    readonly RequestDelegate _next = next ?? (_ => Task.CompletedTask);
+    readonly IPostHogClient? _postHog = postHog;
```
haacked: `suggestion`: The constructor parameters are non-nullable (`RequestDelegate next`, `IPostHogClient postHog`, `PostHogRequestContextOptions options`), but the fields are stored as nullable with `?? (_ => Task.CompletedTask)` / `?? new PostHogRequestContextOptions()` fallbacks and `_postHog` is `IPostHogClient?`. ASP.NET Core's `UseMiddleware<T>` resolves these from DI, so none can be null at runtime. The mismatch makes a future maintainer wonder which contract is real, and the null check inside `TryCaptureException` reads as a guard for a case that can't occur. Tighten the fields to match the parameters:

```suggestion
    readonly RequestDelegate _next = next;
    readonly IPostHogClient _postHog = postHog;
    readonly PostHogRequestContextOptions _options = options;
```

Then the `if (postHog is null)` short-circuit in `TryCaptureException` can also go away.

src/PostHog/PostHogContext.cs
```diff
@@ -0,0 +1,169 @@
...
+internal sealed class PostHogContext
+{
+    static readonly AsyncLocal<PostHogContext?> CurrentContext = new();
+
+    PostHogContext(
+        string? distinctId,
+        string? sessionId,
+        IReadOnlyDictionary<string, object>? properties)
+    {
+        DistinctId = string.IsNullOrEmpty(distinctId) ? null : distinctId;
```
haacked: `suggestion`: `PostHogContext..ctor` and `BeginScope` use `string.IsNullOrEmpty` for distinct-id fallback decisions, while `PostHogTracingHeaders.SanitizeValue` at line 82 uses `string.IsNullOrWhiteSpace`. A request with `X-POSTHOG-DISTINCT-ID: "   "` is sanitized to `null` before reaching `BeginScope` (good), but a programmatic caller doing `PostHogContext.BeginScope(distinctId: "   ")` keeps `"   "` as the distinct id and propagates it to `CaptureCore`. `ResolveIdentity` then treats that whitespace as an explicit user identity instead of personless capture. Behavior should not depend on whether the id arrived via header or in-process. Use `IsNullOrWhiteSpace` in both places, or extract a small `TrimToNull(string?)` helper and use it consistently.

src/PostHog/PostHogContext.cs
```diff
@@ -0,0 +1,169 @@
+using System.Threading;
+using PostHog.Api;
+
+namespace PostHog;
+
+/// <summary>
+/// Request-local PostHog context that is applied to captures in the current async execution flow.
+/// </summary>
+internal sealed class PostHogContext
```
haacked, line 9: `question`: `PostHogContext` and `PostHogContextHelper` are `internal`, so the only way to enter a request scope is the ASP.NET Core middleware. Other PostHog SDKs expose request-scope as public API for non-web hosts: posthog-python (`posthog/contexts.py`) and posthog-node (`packages/node/src/extensions/context/context.ts`) both ship public scope types so background jobs, Worker Services, gRPC servers, and generic hosts can wrap work in a context. Was the choice to keep `PostHogContext` internal driven by API-surface conservatism, or is the long-term plan to add a public scope helper later? If the latter, worth a one-line follow-up tracking issue so a non-ASP.NET consumer doesn't have to wait until they ask.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextExtensions.cs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// <summary>
+    /// Captures an event using the current request context distinct ID, or as a personless event if none is set.
+    /// </summary>
+    /// <param name="client">The <see cref="IPostHogClient"/>.</param>
+    /// <param name="eventName">Human friendly name of the event.</param>
+    /// <param name="properties">Optional: The properties to send along with the event.</param>
+    /// <returns><c>true</c> if the event was successfully enqueued. Otherwise <c>false</c>.</returns>
+    public static bool Capture(
```
haacked, line 38: `question`: Following up on @dustinbyrne's note about `Capture("id", null)` ambiguity (comment 3197019548). Confirming the collision still occurs in the current head SHA: with both `PostHog` and `PostHog.AspNetCore` referenced (the typical case for `UsePostHogRequestContext` users), `posthog.Capture("event-name", null)` and `posthog.CaptureException(ex, null)` fail with `error CS0121` because `null` matches both `string` and `Dictionary<string, object>?`.

The candidates are this file's `Capture(IPostHogClient, string, Dictionary<string, object>?)` and `CaptureExtensions.Capture(IPostHogClient, string, string)` (both in `namespace PostHog`). Workarounds compile cleanly: `client.Capture("event-name", (Dictionary<string, object>?)null)`, `default(Dictionary<string, object>?)`, or `properties: null`. Existing code that wrote `.Capture("distinct-id", null)` (passing `null` for the legacy event-name arg) also fails. Confining the new overloads to `PostHog.AspNetCore` narrowed the collision but didn't remove it.

Worth considering moving these into a non-`PostHog` namespace (e.g. `PostHog.AspNetCore` or `PostHog.RequestContext`) so the new and legacy overloads aren't in the same overload set unless the caller imports both, which would close the gap entirely.

tests/UnitTests.AspNetCore/PostHogRequestContextExtensionsTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        var request = flagsHandler.ReceivedRequests.Single();
+        var body = await request.Content!.ReadAsStringAsync();
+        using var doc = JsonDocument.Parse(body);
+        Assert.Equal("context-user", doc.RootElement.GetProperty("distinct_id").GetString());
+    }
+
+    [Fact]
+    public async Task MissingContextDistinctIdLogsWarningOnceAndAvoidsHttpCalls()
```
haacked, line 30: `suggestion`: `MissingContextDistinctIdLogsWarningOnceAndAvoidsHttpCalls` only exercises `EvaluateFlagsAsync`, but the once-only warning was also added to `IsFeatureEnabledAsync`/`GetFeatureFlagAsync` (`PostHogClient.cs:506-511`) and `GetAllFeatureFlagsAsync` (`:990-995`). A regression that drops the warning or the no-HTTP guard from either of those would ship clean. Add coverage for both:

```csharp
[Fact]
public async Task GetAllFeatureFlagsWithoutContextLogsWarningAndAvoidsHttp()
{
    var flagsHandler = container.FakeHttpMessageHandler.AddFlagsResponse("""{"featureFlags": {"flag-a": true}}""");

    var first = await client.GetAllFeatureFlagsAsync(distinctId: null);
    var second = await client.GetAllFeatureFlagsAsync(distinctId: null);

    Assert.Empty(first);
    Assert.Empty(second);
    Assert.Empty(flagsHandler.ReceivedRequests);
    Assert.Single(
        container.FakeLoggerProvider.GetAllEvents(minimumLevel: LogLevel.Warning),
        e => (e.Message ?? string.Empty).Contains("distinctId is required", StringComparison.Ordinal));
}
```

…and analogous coverage for `IsFeatureEnabledAsync`.

tests/UnitTests/PostHogContextTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                Assert.Null(PostHogContext.Current?.DistinctId);
+                Assert.Null(PostHogContext.Current?.SessionId);
+                Assert.False(PostHogContext.Current?.Properties.ContainsKey("outer"));
+                Assert.Equal(true, PostHogContext.Current?.Properties["fresh"]);
+            }
+        }
+
+        Assert.Null(PostHogContext.Current);
+    }
```
haacked, line 36: `suggestion`: `NestedScopesInheritUnlessFresh` covers an inner scope inheriting from outer (no inner distinctId) and an inner `fresh: true` discarding outer, but never the third path in `BeginScope` where a nested scope passes a non-null distinctId and *overrides* the parent. That branch (line 71, `string.IsNullOrEmpty(distinctId) ? baseContext?.DistinctId : distinctId`) has no test today, and neither does the empty-string-falls-back-to-parent case. A regression that swaps the conditional or treats `""` as a valid override would not be caught.

```csharp
[Fact]
public void InnerScopeDistinctIdOverridesParent()
{
    using (PostHogContext.BeginScope(distinctId: "outer-user", fresh: true))
    using (PostHogContext.BeginScope(distinctId: "inner-user"))
    {
        Assert.Equal("inner-user", PostHogContext.Current?.DistinctId);
    }
}

[Fact]
public void InnerScopeEmptyDistinctIdInheritsParent()
{
    using (PostHogContext.BeginScope(distinctId: "outer-user", fresh: true))
    using (PostHogContext.BeginScope(distinctId: ""))
    {
        Assert.Equal("outer-user", PostHogContext.Current?.DistinctId);
    }
}
```

tests/UnitTests.AspNetCore/PostHogRequestContextMiddlewareTests.cs
```diff
@@ -0,0 +1,362 @@
...
+        second.Request.Headers[PostHogTracingHeaders.SessionId] = "session-b";
+
+        await Task.WhenAll(middleware.InvokeAsync(first), middleware.InvokeAsync(second));
+
+        Assert.Equal(("user-a", "session-a"), results["/first"]);
+        Assert.Equal(("user-b", "session-b"), results["/second"]);
+        Assert.Null(PostHogContext.Current);
+    }
+
+    [Fact]
```
haacked: `suggestion`: `CapturesUnhandledExceptionsWithRequestContextAndRethrows` asserts `$response_status_code: 503` for a 503 response, but never exercises the fallback: when the response status code is below 400, `GetExceptionStatusCode` returns 500. Today nothing pins that mapping. If someone changed the threshold or the default, an exception thrown during a still-200 response would silently report `$response_status_code: 200`. Convert to a `[Theory]` covering both branches:

```csharp
[Theory]
[InlineData(200, 500)]
[InlineData(399, 500)]
[InlineData(400, 400)]
[InlineData(503, 503)]
public async Task ExceptionStatusCodeUses500FallbackBelow400(int responseStatusCode, int expected)
{
    var requestHandler = container.FakeHttpMessageHandler.AddBatchResponse();
    var httpContext = CreateHttpContext();
    httpContext.Response.StatusCode = responseStatusCode;

    var middleware = CreateMiddleware(
        _ => throw new InvalidOperationException("boom"),
        client,
        options => options.CaptureExceptions = true);

    await Assert.ThrowsAsync<InvalidOperationException>(() => middleware.InvokeAsync(httpContext));
    await client.FlushAsync();

    using var doc = JsonDocument.Parse(requestHandler.GetReceivedRequestBody(indented: false));
    var properties = doc.RootElement.GetProperty("batch")[0].GetProperty("properties");
    Assert.Equal(expected, properties.GetProperty("$response_status_code").GetInt32());
}
```

tests/UnitTests/PostHogContextTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        var batchItem = document.RootElement.GetProperty("batch")[0];
+        Assert.Equal("explicit-user", batchItem.GetProperty("distinct_id").GetString());
+        var properties = batchItem.GetProperty("properties");
+        Assert.Equal("explicit-user", properties.GetProperty("distinct_id").GetString());
+        Assert.Equal("explicit-session", properties.GetProperty("$session_id").GetString());
+        Assert.Equal("explicit-value", properties.GetProperty("shared").GetString());
+        Assert.Equal("context-only-value", properties.GetProperty("context-only").GetString());
+    }
+
+    [Theory]
```
haacked, line 144: `nit`: The `[Theory]` covers `null` (no override) and `true` (explicit override), but skips the `false` row. The contract is "if the caller passed `$process_person_profile` we don't overwrite it", and `false` is the only value that genuinely differs from the personless default. Add the row to lock that down:

```suggestion
    [Theory]
    [InlineData(null, false)]
    [InlineData(true, true)]
    [InlineData(false, false)]
    public void PersonlessContextSetsProcessPersonProfile(bool? explicitOverride, bool expectedValue)
```

tests/UnitTests.AspNetCore/PostHogRequestContextMiddlewareTests.cs
```diff
@@ -0,0 +1,362 @@
...
+        using var document = JsonDocument.Parse(requestHandler.GetReceivedRequestBody(indented: false));
+        var batchItem = document.RootElement.GetProperty("batch")[0];
+        Assert.True(Guid.TryParse(batchItem.GetProperty("distinct_id").GetString(), out _));
+        Assert.NotEqual("authenticated-user", batchItem.GetProperty("distinct_id").GetString());
+        var properties = batchItem.GetProperty("properties");
+        Assert.False(properties.GetProperty("$process_person_profile").GetBoolean());
+        Assert.False(properties.TryGetProperty("$session_id", out _));
+    }
+
+    [Fact]
```
haacked, line 228: `suggestion`: The middleware tests always set `RemoteIpAddress` and a non-empty `Scheme`/`Host` via `CreateHttpContext`, so the two short-circuit branches in `PostHogTracingHeaders` (`AddIfPresent` skipping null IP, `GetCurrentUrl` returning null when scheme is empty or host has no value) are unexercised. A regression that started writing `"$ip": null` or `"$current_url": "://"` would slip past CI.

```csharp
[Fact]
public async Task OmitsCurrentUrlAndIpWhenSchemeOrConnectionAreMissing()
{
    var requestHandler = container.FakeHttpMessageHandler.AddBatchResponse();
    var httpContext = new DefaultHttpContext();
    httpContext.Request.Method = "GET";
    httpContext.Request.Path = "/api/test";
    // Scheme/Host left unset, RemoteIpAddress left null.

    var middleware = CreateMiddleware(_ => { client.Capture("event"); return Task.CompletedTask; }, client);
    await middleware.InvokeAsync(httpContext);
    await client.FlushAsync();

    using var doc = JsonDocument.Parse(requestHandler.GetReceivedRequestBody(indented: false));
    var properties = doc.RootElement.GetProperty("batch")[0].GetProperty("properties");
    Assert.False(properties.TryGetProperty("$current_url", out _));
    Assert.False(properties.TryGetProperty("$ip", out _));
}
```

src/PostHog.AspNetCore/Tracing/PostHogTracingHeaders.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            properties[key] = value;
+        }
+    }
+
+    static string? GetCurrentUrl(HttpRequest request)
+    {
+        if (string.IsNullOrEmpty(request.Scheme) || !request.Host.HasValue)
+        {
+            return null;
```
haacked, line 112: `suggestion`: `GetCurrentUrl` now appends `request.QueryString.ToUriComponent()` to `$current_url` so PostHog gets URL parity with the analytics dashboards (greptile P1 in comment 3197182739). The tradeoff is that every server capture now ships the full query string verbatim, including OAuth `code=`, password-reset tokens, signed-URL params, and any other secret a redirect can land on the server. Other server SDKs sidestep this by either dropping the query string for `$current_url` or providing an opt-out. Two reasonable options:

1. Strip well-known sensitive keys (`code`, `state`, `token`, `access_token`, `id_token`, `password`, etc.) before stringifying.
2. Add a `PostHogRequestContextOptions.IncludeQueryString` toggle (default `true` for parity, `false` for callers who handle auth callbacks).

At minimum, document the leak surface in the option's doc comment so a careful reader sees it without leaving the IDE.

src/PostHog.AspNetCore/Tracing/PostHogTracingHeaders.cs
```diff
@@ -0,0 +1,139 @@
...
+            var sessionId = options.UseTracingHeaders
+                ? SanitizeHeaderValue(request.Headers[SessionId])
+                : null;
+
+            var properties = new Dictionary<string, object>();
+            AddIfPresent(properties, PostHogRequestPropertyNames.CurrentUrl, GetCurrentUrl(request));
+            AddIfPresent(properties, PostHogRequestPropertyNames.RequestMethod, request.Method);
+            AddIfPresent(properties, PostHogRequestPropertyNames.RequestPath, GetRequestPath(request));
+            AddIfPresent(properties, PostHogRequestPropertyNames.UserAgent, SanitizeHeaderValue(request.Headers[HeaderNames.UserAgent]));
+            AddIfPresent(properties, PostHogRequestPropertyNames.Ip, httpContext.Connection.RemoteIpAddress?.ToString());
```
haacked: `question`: `httpContext.Connection.RemoteIpAddress?.ToString()` becomes `$ip` unconditionally. The README warns the user must wire `UseForwardedHeaders` upstream of this middleware, but when they don't (or order them wrong), `RemoteIpAddress` is the load-balancer or pod-network peer (`10.0.x.x`, the K8s service IP) and every captured event ships an internal cluster IP to PostHog. Was that intentional (catch the misconfiguration in the dashboard data) or worth surfacing in the option's doc comment so a careful reader sees the failure mode without leaving the IDE?

haacked (commented):
Solid PR overall. A handful of non-blocking suggestions inline; the highest-leverage ones are the duplicate `$session_id` constant, the empty-string divergence across the three flag entrypoints, and the query-string secrets surface in `$current_url`.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextExtensions.cs
dustinbyrne: Good callout. I’m going to keep the parameterless request-context flag overloads because they’re useful when an app intentionally opts into request-context identity by installing the middleware and using context-aware call sites. They also help maintain stable frontend/backend flag evaluations out of the box for teams that want consistency within a request/session. Explicit distinct IDs still override the context. I've strengthened the docs to say that header-derived identity is client-controlled analytics context, not authentication or authorization context, and security-sensitive server branching should pass an authenticated distinct ID explicitly.

I agree the client-controlled nature needs to be prominent, but I don’t think we should flip the default. The opt-in is at the middleware and call-site level: apps install `UsePostHogRequestContext()`, then choose context-aware calls by omitting `distinctId` / using parameterless helpers. Explicit distinct IDs override the context. I’ll keep the default and add docs warning that tracing headers are analytics context, not auth, with `UseTracingHeaders = false` remaining available for metadata-only mode.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextMiddlewareExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <param name="configure">Optional configuration for tracing headers and exception capture.</param>
+    /// <returns>The passed in <see cref="IApplicationBuilder" />.</returns>
+    public static IApplicationBuilder UsePostHogRequestContext(
+        this IApplicationBuilder app,
+        Action<PostHogRequestContextOptions>? configure = null)
+    {
+        if (app is null)
+        {
+            return app!;
+        }
```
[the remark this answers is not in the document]
dustinbyrne: Agreed. This now fails fast with `ArgumentNullException.ThrowIfNull(app)`; returning null would mask a startup/configuration bug.

src/PostHog.AspNetCore/Tracing/PostHogTracingHeaders.cs
dustinbyrne: Thanks. I’m going to keep the simpler implementation for now; the suggested `Span` version changes trimming/truncation behavior for long values with leading whitespace. I’d prefer to optimize this with benchmark/profile evidence.

src/PostHog/PostHogContext.cs
dustinbyrne: Intentional for this PR. I kept the context type internal to limit public API surface while validating the ASP.NET middleware. I agree Worker Services/gRPC/background jobs need a public scope helper; We can track that as a follow-up rather than expose `PostHogContext` directly here.

src/PostHog/Capture/CaptureExtensions.cs
dustinbyrne: Moved the extension into the `PostHog.AspNetCore` namespace. This is fully backwards compatible.

src/PostHog.AspNetCore/Tracing/PostHogRequestPropertyNames.cs
dustinbyrne: Addressed in c0ec277.

src/PostHog/PostHogClient.cs
dustinbyrne: Addressed in fb1071a; added coverage in 3f7b5e4.

src/PostHog/PostHogContext.cs
dustinbyrne: Addressed in 81652b1.

src/PostHog/PostHogContext.cs
dustinbyrne: Addressed in 9d2257e.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextExtensions.cs
dustinbyrne: Documented in 7d0b904, 7f55f30, and ff68b70.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextMiddleware.cs
dustinbyrne: Addressed in b419924.

src/PostHog/PostHogContext.cs
dustinbyrne: Addressed in fb1071a; added programmatic whitespace coverage in 4ded41f.

src/PostHog.AspNetCore/Tracing/PostHogRequestContextExtensions.cs
dustinbyrne: Good catch. Addressed in 64216d5 by moving the ASP.NET request-context convenience extensions out of the root `PostHog` namespace, so existing `using PostHog;` callers don't get the new nullable overloads in their overload set.

tests/UnitTests.AspNetCore/PostHogRequestContextExtensionsTests.cs
dustinbyrne: Addressed in d56a2dc, b46c7da, and 3f7b5e4.

tests/UnitTests/PostHogContextTests.cs
dustinbyrne: Addressed in 0087904.

tests/UnitTests.AspNetCore/PostHogRequestContextMiddlewareTests.cs
dustinbyrne: Addressed in 720bc04.

tests/UnitTests/PostHogContextTests.cs
dustinbyrne: Addressed in 1351aaa.

tests/UnitTests.AspNetCore/PostHogRequestContextMiddlewareTests.cs
dustinbyrne: Addressed in d74605f.

src/PostHog.AspNetCore/Tracing/PostHogTracingHeaders.cs
dustinbyrne: Addressed in 5aedf52 and 58bad2b. `$current_url` now drops the query string entirely; coverage added in 94f45f7.

src/PostHog.AspNetCore/Tracing/PostHogTracingHeaders.cs
dustinbyrne: Intentional to use ASP.NET Core's normalized `RemoteIpAddress`, but agreed the proxy/order failure mode should be visible in IDE docs too. Addressed in 3b47030.

marandaneto:
@haacked did the review already so skipping until its good to go, happy to do a final review if needed otherwise his approval should be enough!

src/PostHog.AspNetCore/Tracing/PostHogRequestContextExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// Captures an event using the current request context distinct ID, or as a personless event if none is set.
+    /// </summary>
+    /// <param name="client">The <see cref="IPostHogClient"/>.</param>
+    /// <param name="eventName">Human friendly name of the event.</param>
+    /// <returns><c>true</c> if the event was successfully enqueued. Otherwise <c>false</c>.</returns>
+    public static bool Capture(
+        this IPostHogClient client,
+        string eventName)
+    {
+        var checkedClient = NotNull(client);
```
marandaneto, line 21: this will crash the app, better to just return false instead, check the other callsites

marandaneto (approved):

src/PostHog.AspNetCore/Tracing/PostHogRequestContextExtensions.cs
dustinbyrne: Maybe worth taking a broader look at this pattern in this SDK as it's pretty ubiquitous.

The blast radius where a null client is possible is limited given that the client is provided via ASP.NET dependency injection, and if the client wasn't registered, it'd be an error ahead of this null check.

E.g.:
```
InvalidOperationException: Unable to resolve service for type 'PostHog.IPostHogClient' while attempting to activate 'PostHog.PostHogRequestContextMiddleware'
```

marandaneto: Gotcha

PR #206, opened 2026-05-20
fix: reject leading-zero semver values in local evaluation

dmarticus:
## Summary

Per semver 2.0.0 §2, numeric identifiers MUST NOT include leading zeros. Previously, `SemanticVersion.TryParse` and `TryParseWildcard` used `int.TryParse` directly, which accepted strings like `"01.02.03"` and `"1.07.3"` and quietly converted them to `1.2.3` / `1.7.3`. That meant feature flag targeting conditions could match on invalid semver inputs, producing incorrect rollout results.

This brings posthog-dotnet in line with the matching fixes in posthog-python (#601) and posthog-go (#200).

## Changes

- Added `TryParseSemverNumeric` helper in `SemanticVersion` that rejects empty input, non-digit characters, and any value with a leading zero (except literal `"0"`).
- Used the helper inside `TryParse` (major/minor/patch) and `TryParseWildcard` (each numeric segment).
- Removed the now-redundant negative-component check from `TryParse` — non-digits are rejected up front.
- Inverted the existing leading-zero acceptance test cases and added new rejection cases covering all semver operators (`semver_eq`, `semver_neq`, `semver_gt`, `semver_gte`, `semver_lt`, `semver_lte`, `semver_tilde`, `semver_caret`, `semver_wildcard`) for both override values and flag values.
- Literal `"0"` components (e.g. `"0.1.0"`, `"1.0.0"`, `"0.0.0"`) remain valid.

Invalid inputs cause `TryParse` to return `false`, which causes the higher-level evaluator to raise `InconclusiveMatchException` for that condition — matching the existing behavior for any other unparseable semver value.

## Test plan

- [x] `dotnet test -f net8.0` — all previously passing tests still pass (2 pre-existing failures in `PostHogClientTests.TheCaptureExceptionMethod` are unrelated and reproduce on `main`)
- [x] New `SemanticVersionTests` cases cover leading zeros in major/minor/patch, multi-zero, and v-prefixed inputs
- [x] New `TryParseWildcard` cases reject `"01.*"`, `"1.02.*"`, `"01.2.*"`, `"01"`, `"01.2"`, `"1.02"`
- [x] `TheSemverOperators` cases verify rejection across every semver operator for both override and flag values
- [x] Literal-zero components (`0.0.0`, `0.1.0`, `1.0.0`, `1.2.0`) still parse and match

### DIFF
--- a/src/PostHog/Library/SemanticVersion.cs
+++ b/src/PostHog/Library/SemanticVersion.cs
@@ -112,21 +112,21 @@ internal readonly record struct SemanticVersion : IComparable<SemanticVersion>
         var parts = trimmed.Split('.');
 
         // Parse major (required)
-        if (!TryParseSemverNumeric(parts[0], out var major))
+        if (!TryParseSemverNumericIdentifier(parts[0], out var major))
         {
             return false;
         }
 
         // Parse minor (optional, defaults to 0)
         var minor = 0;
-        if (parts.Length > 1 && !string.IsNullOrEmpty(parts[1]) && !TryParseSemverNumeric(parts[1], out minor))
+        if (parts.Length > 1 && !string.IsNullOrEmpty(parts[1]) && !TryParseSemverNumericIdentifier(parts[1], out minor))
         {
             return false;
         }
 
         // Parse patch (optional, defaults to 0)
         var patch = 0;
-        if (parts.Length > 2 && !string.IsNullOrEmpty(parts[2]) && !TryParseSemverNumeric(parts[2], out patch))
+        if (parts.Length > 2 && !string.IsNullOrEmpty(parts[2]) && !TryParseSemverNumericIdentifier(parts[2], out patch))
         {
             return false;
         }
@@ -135,15 +135,20 @@ internal readonly record struct SemanticVersion : IComparable<SemanticVersion>
         return true;
     }
 
-    // Semver 2.0.0 §2: numeric identifiers MUST NOT include leading zeros.
-    static bool TryParseSemverNumeric(string part, out int value)
+    static bool TryParseSemverNumericIdentifier(string part, out int value)
     {
         value = 0;
-        if (string.IsNullOrEmpty(part) || (part.Length > 1 && part[0] == '0'))
+        if (string.IsNullOrEmpty(part))
+        {
+            return false;
+        }
+        // Semver 2.0.0 §2: literal "0" is allowed; "01", "00", "001" are not.
+        if (part.Length > 1 && part[0] == '0')
         {
             return false;
         }
         // NumberStyles.None + InvariantCulture rejects signs, whitespace, and non-ASCII digits.
+        // The sign rejection is what lets TryParse drop the explicit negative-component check.
         return int.TryParse(part, NumberStyles.None, CultureInfo.InvariantCulture, out value);
     }
 
@@ -260,7 +265,7 @@ internal readonly record struct SemanticVersion : IComparable<SemanticVersion>
         {
             // "X" pattern - treat as "X.*"
             // Bare wildcards like "*" and non-numeric values are invalid
-            if (!TryParseSemverNumeric(parts[0], out var major))
+            if (!TryParseSemverNumericIdentifier(parts[0], out var major))
             {
                 return false;
             }
@@ -272,7 +277,7 @@ internal readonly record struct SemanticVersion : IComparable<SemanticVersion>
         else if (parts.Length == 2)
         {
             // "X.Y" or "X.*" pattern
-            if (!TryParseSemverNumeric(parts[0], out var major))
+            if (!TryParseSemverNumericIdentifier(parts[0], out var major))
             {
                 return false;
             }
@@ -285,7 +290,7 @@ internal readonly record struct SemanticVersion : IComparable<SemanticVersion>
                 return true;
             }
 
-            if (!TryParseSemverNumeric(parts[1], out var minor))
+            if (!TryParseSemverNumericIdentifier(parts[1], out var minor))
             {
                 return false;
             }
@@ -298,12 +303,12 @@ internal readonly record struct SemanticVersion : IComparable<SemanticVersion>
         else if (parts.Length >= 3)
         {
             // "X.Y.Z" or "X.Y.*" pattern
-            if (!TryParseSemverNumeric(parts[0], out var major))
+            if (!TryParseSemverNumericIdentifier(parts[0], out var major))
             {
                 return false;
             }
 
-            if (!TryParseSemverNumeric(parts[1], out var minor))
+            if (!TryParseSemverNumericIdentifier(parts[1], out var minor))
             {
                 return false;
             }
--- a/src/PostHog/Library/SemanticVersion.cs
+++ b/src/PostHog/Library/SemanticVersion.cs
@@ -47,6 +47,7 @@ internal readonly record struct SemanticVersion : IComparable<SemanticVersion>
     /// 5. Default missing components to 0 (e.g., "1.2" → (1, 2, 0), "1" → (1, 0, 0))
     /// 6. Ignore extra components beyond the third (e.g., "1.2.3.4" → (1, 2, 3))
     /// 7. Return false for truly invalid input (empty string, non-numeric parts, leading dot)
+    /// 8. Reject components with leading zeros per semver 2.0.0 §2 (e.g., "01.02.03")
     /// </remarks>
     public static bool TryParse(string? value, [NotNullWhen(returnValue: true)] out SemanticVersion? version)
     {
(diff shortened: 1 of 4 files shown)


tests/UnitTests/Library/SemanticVersionTests.cs
```diff
@@ -139,6 +141,13 @@ public void ParsesLeadingZeros(string input, int expectedMajor, int expectedMino
     [InlineData("1.-2.3")]    // Negative minor
     [InlineData("-1.2.3")]    // Negative major
     [InlineData("1.2.-3")]    // Negative patch
+    [InlineData("01.02.03")]  // Leading zeros (all components)
+    [InlineData("001.002.003")] // Multiple leading zeros
+    [InlineData("1.07.3")]    // Leading zero in minor
+    [InlineData("1.2.03")]    // Leading zero in patch
+    [InlineData("01.2.3")]    // Leading zero in major
+    [InlineData("00.0.0")]    // Leading zero on a zero major
+    [InlineData("v01.2.3")]   // Leading zero with v-prefix
```
haacked, line 150: `suggestion`: The new `ReturnsFalseForInvalidInput` rows cover bare leading-zero triples and `"v01.2.3"`, but nothing pins the contract that the leading-zero check survives the upstream pre-release, build-metadata, and outer-whitespace strips. If a future change reorders `TryParse` so the leading-zero check runs before those strips, today's tests still pass while `"01.2.3-alpha"` silently goes back to being accepted as `1.2.3`. A few rows lock the composition in:

```csharp
[InlineData("01.2.3-alpha")]      // Leading zero + pre-release
[InlineData("01.2.3+build")]      // Leading zero + build metadata
[InlineData("  01.2.3  ")]        // Leading zero + outer whitespace
```

tests/UnitTests/Library/SemanticVersionTests.cs
```diff
@@ -139,6 +141,13 @@ public void ParsesLeadingZeros(string input, int expectedMajor, int expectedMino
     [InlineData("1.-2.3")]    // Negative minor
     [InlineData("-1.2.3")]    // Negative major
     [InlineData("1.2.-3")]    // Negative patch
+    [InlineData("01.02.03")]  // Leading zeros (all components)
```
haacked, line 144: `suggestion`: `TryParseSemverNumeric` relies on `NumberStyles.None + InvariantCulture` to reject signs, embedded whitespace, and non-ASCII digits, plus on `int.TryParse` returning false on overflow. The new leading-zero rows exercise the explicit `part[0] == '0'` guard; the rest of the helper's contract is only covered by the pre-existing `-1.2.3`/`1.-2.3` sign cases. If someone later swaps in `int.Parse(part)` or relaxes the styles, every present test still passes. Three rows close the gap:

```csharp
[InlineData("1. 2.3")]              // Embedded whitespace in minor (NumberStyles.None)
[InlineData("١.2.3")]              // Arabic-Indic digit in major (InvariantCulture)
[InlineData("99999999999999.0.0")]  // Overflows int.MaxValue
```

tests/UnitTests/Features/LocalEvaluatorTests.cs
```diff
@@ -2358,11 +2365,22 @@ public void ThrowsInconclusiveMatchExceptionForInvalidOverrideVersion(string ove
...
     // Invalid version in filter value - should throw InconclusiveMatchException
-    [InlineData("not-a-version")]
-    [InlineData("")]
-    [InlineData("abc.def.ghi")]
-    [InlineData(".1.2.3")]
-    public void ThrowsInconclusiveMatchExceptionForInvalidFilterVersion(string filterValue)
+    [InlineData("not-a-version", ComparisonOperator.SemverEquals)]
+    [InlineData("", ComparisonOperator.SemverEquals)]
+    [InlineData("abc.def.ghi", ComparisonOperator.SemverEquals)]
+    [InlineData(".1.2.3", ComparisonOperator.SemverEquals)]
```
haacked, line 2371: `suggestion`: The signature change to this test opens the door to running each invalid value under every operator, but the four pre-existing rows (`"not-a-version"`, `""`, `"abc.def.ghi"`, `".1.2.3"`) only get `SemverEquals`. The override-version twin test above already pairs `"not-a-version"` and `""` with `SemverGreaterThan`. Mirror that here so the non-equals operators stay covered for non-leading-zero malformed input:

```csharp
[InlineData("not-a-version", ComparisonOperator.SemverGreaterThan)]
[InlineData("", ComparisonOperator.SemverGreaterThan)]
[InlineData("abc.def.ghi", ComparisonOperator.SemverTilde)]
[InlineData(".1.2.3", ComparisonOperator.SemverCaret)]
```

src/PostHog/Library/SemanticVersion.cs
```diff
@@ -110,33 +112,39 @@ public static bool TryParse(string? value, [NotNullWhen(returnValue: true)] out
...
+    {
+        value = 0;
+        if (string.IsNullOrEmpty(part) || (part.Length > 1 && part[0] == '0'))
         {
             return false;
         }
-
-        version = new SemanticVersion(major, minor, patch);
-        return true;
+        // NumberStyles.None + InvariantCulture rejects signs, whitespace, and non-ASCII digits.
```
haacked, line 150: `suggestion`: The deletion of the explicit `if (major < 0 || minor < 0 || patch < 0)` check is safe only because `NumberStyles.None` rejects sign characters during parse. That coupling is load-bearing and invisible. The existing comment names the behavior; spelling out the dependency keeps the next person from "tightening" to `NumberStyles.Integer` and silently re-opening negative-component acceptance.

```suggestion
        // NumberStyles.None + InvariantCulture rejects signs, whitespace, and non-ASCII digits.
        // The sign rejection is what lets TryParse drop the explicit negative-component check above.
```

tests/UnitTests/Library/SemanticVersionTests.cs
```diff
@@ -470,6 +479,12 @@ public void ParsesImplicitMinorWildcard(string pattern, string expectedLower, st
     [InlineData("1.2.3.*")]  // Too specific
     [InlineData(".1.*")]     // Leading dot
     [InlineData("abc.*")]    // Non-numeric
+    [InlineData("01.*")]     // Leading zero in major
+    [InlineData("1.02.*")]   // Leading zero in minor
+    [InlineData("01.2.*")]   // Leading zero in major (X.Y.*)
+    [InlineData("01")]       // Leading zero in implicit major
+    [InlineData("01.2")]     // Leading zero in implicit X.Y
+    [InlineData("1.02")]     // Leading zero in implicit minor
```
haacked, line 493: `suggestion`: `ReturnsFalseForInvalidPatterns` covers leading zeros in major/minor and the implicit `X`/`X.Y` forms but is missing two cases that `ReturnsFalseForInvalidInput` does cover for `TryParse`: a v-prefixed leading zero and a multiple-leading-zero variant. Both paths go through `TryParseSemverNumeric`, so behavior should already be correct, but the symmetry protects against someone later short-circuiting the wildcard path or moving the strip-v logic:

```csharp
[InlineData("v01.*")]    // Leading zero with v-prefix
[InlineData("001.*")]    // Multiple leading zeros
```

src/PostHog/Library/SemanticVersion.cs
```diff
@@ -110,33 +112,39 @@ public static bool TryParse(string? value, [NotNullWhen(returnValue: true)] out
...
-        if (major < 0 || minor < 0 || patch < 0)
+        version = new SemanticVersion(major, minor, patch);
+        return true;
+    }
+
+    // Semver 2.0.0 §2: numeric identifiers MUST NOT include leading zeros.
+    static bool TryParseSemverNumeric(string part, out int value)
+    {
+        value = 0;
+        if (string.IsNullOrEmpty(part) || (part.Length > 1 && part[0] == '0'))
```
haacked: `nit`: This guard packs two independent failure modes into one boolean: empty input and leading zero. A reader has to parse the precedence of `&&` inside `||` and remember the `Length > 1` exception exists for the literal `"0"` case. Splitting reads more like the spec rules in the `<remarks>` block:

```csharp
if (string.IsNullOrEmpty(part)) return false;
// Semver 2.0.0 §2: literal "0" is allowed; "01", "00", "001" are not.
if (part.Length > 1 && part[0] == '0') return false;
return int.TryParse(part, NumberStyles.None, CultureInfo.InvariantCulture, out value);
```

tests/UnitTests/Library/SemanticVersionTests.cs
```diff
@@ -107,10 +107,12 @@ public void IgnoresExtraComponents(string input, int expectedMajor, int expected
     }
 
     [Theory]
-    // Leading zeros are parsed as integers
-    [InlineData("01.02.03", 1, 2, 3)]
-    [InlineData("001.002.003", 1, 2, 3)]
-    public void ParsesLeadingZeros(string input, int expectedMajor, int expectedMinor, int expectedPatch)
+    // Literal "0" components are valid per semver 2.0.0
+    [InlineData("0.0.0", 0, 0, 0)]
```
haacked: `nit`: `[InlineData("0.0.0", 0, 0, 0)]` here is already covered verbatim by `ParsesValidVersions` higher up in the file. Drop this row or swap it for a case that isn't already pinned (e.g., `[InlineData("0.0.1", 0, 0, 1)]`). The other three new rows do pull their weight.

src/PostHog/Library/SemanticVersion.cs
```diff
@@ -110,33 +112,39 @@ public static bool TryParse(string? value, [NotNullWhen(returnValue: true)] out
...
         }
 
-        // Reject negative version components
-        if (major < 0 || minor < 0 || patch < 0)
+        version = new SemanticVersion(major, minor, patch);
+        return true;
+    }
+
+    // Semver 2.0.0 §2: numeric identifiers MUST NOT include leading zeros.
+    static bool TryParseSemverNumeric(string part, out int value)
```
haacked: `nit`: `TryParseSemverNumeric` reads like "parse a full semver string" but actually parses a single numeric identifier. `TryParseSemverNumericIdentifier` or `TryParseSemverComponent` matches the doc comment ("numeric identifiers MUST NOT include leading zeros") more directly. Stylistic.

marandaneto (approved):

marandaneto:
missing changelog entry

haacked (commented):
Looks good to me. Have some non-blocking test case suggestions, but other than that, 🚢

haacked (approved):
:shipit:

PR #209, opened 2026-05-25
fix: include group context in $feature_flag_called dedupe key

### LINKED ISSUE #3651: $feature_flag_called dedupe does not include group context for group-scoped flags
## Summary

In `posthog-node`, `$feature_flag_called` events appear to be deduped by `(distinctId, flag key, response)` only. The group context passed to feature flag evaluation is attached to the captured event, but does not appear to participate in the SDK-level dedupe key.

For group-scoped feature flags / experiments, this can suppress exposure events for a second group when the same user evaluates the same flag and receives the same response in both groups.

I'm not sure if this is intended behavior or if there's a recommended usage pattern we're missing, so filing this as a question / possible bug.

## Why this matters

For a group-scoped experiment, assignment can be based on the group key, and experiment metrics may join exposure events using a group property like `$group_0`.

If the same user visits two different groups and receives the same variant in both, the second group may not get a `$feature_flag_called` event. That means group-level exposure counts and warehouse-backed experiment joins can undercount exposed groups.

## Reproduction

Using `posthog-node@5.34.3`:

```ts
import { PostHog } from 'posthog-node'

const client = new PostHog('ph_test_key', {
  host: 'http://127.0.0.1:9',
  flushAt: 9999,
  flushInterval: 0,
})

const captured: any[] = []
client.capture = (event: any) => captured.push(event)

function fire({
  distinctId,
  response,
  org,
}: {
  distinctId: string
  response: string
  org: string
}) {
  // Calling the internal method here only to isolate the dedupe behavior.
  ;(client as any)._captureFlagCalledEventIfNeeded({
    distinctId,
    key: 'group-scoped-experiment',
    response,
    groups: { org },
    properties: {
      $feature_flag: 'group-scoped-experiment',
      $feature_flag_response: response,
      '$feature/group-scoped-experiment': response,
    },
  })
}

fire({ distinctId: 'user_123', response: 'control', org: 'org_A' })
fire({ distinctId: 'user_123', response: 'control', org: 'org_B' })
fire({ distinctId: 'user_123', response: 'test', org: 'org_B' })
fire({ distinctId: 'user_456', response: 'control', org: 'org_B' })

console.log(
  captured.map(({ distinctId, event, groups, properties }) => ({
    distinctId,
    event,
    groups,
    response: properties.$feature_flag_response,
  })),
)
```

## Actual behavior

The second call is skipped:

```json
[
  {
    "distinctId": "user_123",
    "event": "$feature_flag_called",
    "groups": { "org": "org_A" },
    "response": "control"
  },
  {
    "distinctId": "user_123",
    "event": "$feature_flag_called",
    "groups": { "org": "org_B" },
    "response": "test"
  },
  {
    "distinctId": "user_456",
    "event": "$feature_flag_called",
    "groups": { "org": "org_B" },
    "response": "control"
  }
]
```

`user_123 + control + org_B` is not emitted because `user_123 + group-scoped-experiment + control` was already emitted for `org_A`.

## Expected / question

For group-scoped flags, should the dedupe key include group context, at least the aggregation group used by the flag?

For example:

```ts
distinctId + flag key + response + relevant group type/key
```

Or is the intended usage for group-scoped experiments to pass the group key as `distinctId`, e.g. `evaluateFlags(orgId, { groups: { org: orgId } })`?

## Relevant SDK code

In current `posthog-js` main, the dedupe key appears to ignore groups:

```ts
const { distinctId, key, response, groups, disableGeoip, properties } = params
const featureFlagReportedKey = `${key}_${response}`

if (
  distinctId in this.distinctIdHasSentFlagCalls &&
  this.distinctIdHasSentFlagCalls[distinctId].includes(featureFlagReportedKey)
) {
  return
}
```

`groups` is then attached only when the event is captured:

```ts
this.capture({
  distinctId,
  event: '$feature_flag_called',
  properties,
  groups,
  disableGeoip,
})
```

## Environment

- Observed with `posthog-node@5.34.3`
- Current npm latest checked: `posthog-node@5.34.10`
- Current `PostHog/posthog-js` main appears to have the same dedupe shape

## Context

We hit this while instrumenting a group-scoped A/A experiment where:

- the flag is scoped to an `org` group
- the exposure event is `$feature_flag_called`
- the experiment metric joins exposure using the org group key
- the same authenticated user can switch between multiple orgs in the same app/server process

The local workaround is to use a fresh PostHog client for this specific exposure path so the SDK-level dedupe cache does not span orgs, while keeping a singleton client for other server-side usage.

gustavohstrassburger:

In `PostHogClient.TryCaptureDedupedFeatureFlagCalledEvent` (`src/PostHog/PostHogClient.cs`), the per-`distinctId` `MemoryCache` key only included `(distinctId, featureKey, cacheKeyValue)`. For group-scoped flags this meant that when the same user was evaluated under a different group, no new `$feature_flag_called` event was fired — causing per-group exposure undercount for experiments scoped to a group key.

Mirrors the posthog-node fix in PostHog/posthog-js#3658 (which closes PostHog/posthog-js#3651). Both SDKs share the same dedupe shape, so backend evaluation needs the same change.

Added three xUnit tests in `tests/UnitTests/Features/FeatureFlagsTests.cs`:

- `CapturesSeparateFeatureFlagCalledEventsPerGroupContext` — same user, same flag, two different group contexts → two events fired.
- `DedupesFeatureFlagCalledAcrossRepeatedCallsUnderSameGroupContext` — repeated access under the same group → one event.
- `DedupesFeatureFlagCalledAcrossGroupInsertionOrder` — same groups added in different insertion order → one event (canonicalized via `OrderBy(GroupType, StringComparer.Ordinal)`).

`dotnet test --filter "FullyQualifiedName~Feature"` is green: 322 tests pass on net8.0, 312 on netcoreapp3.1.

## :pencil: Checklist

### If releasing new changes

---

### DIFF
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -733,7 +733,7 @@ public sealed class PostHogClient : IPostHogClient
 
         _featureFlagCalledEventCache.GetOrCreate(
             key: (distinctId, featureKey, cacheKeyValue, groupsCacheKey),
-            // This factory only runs when the (distinct id, key, value, groups) tuple is not yet cached.
+            // This factory only runs when the (distinctId, featureKey, cacheKeyValue, groupsCacheKey) tuple is not yet cached.
             // Group context is included so group-scoped flags fire a separate event for each group
             // a user is evaluated under, instead of dedup-ing across groups.
             factory: cacheEntry =>
@@ -765,8 +765,8 @@ public sealed class PostHogClient : IPostHogClient
 
     // Build a canonical string for the groups collection that is order-independent
     // so two equal sets of groups always produce the same dedup cache key.
-    // Returns the empty string when no groups were passed, which keeps the legacy
-    // "no groups" dedupe shape for callers that don't pass a GroupCollection.
+    // Returns the empty string when no groups were passed, preserving the
+    // no-groups dedupe shape for callers that don't pass a GroupCollection.
     static string CanonicalGroupsCacheKey(GroupCollection? groups)
     {
         if (groups is null || groups.Count == 0)
@@ -774,9 +774,8 @@ public sealed class PostHogClient : IPostHogClient
             return string.Empty;
         }
         var pairs = groups
-            .Select(g => (g.GroupType, g.GroupKey))
-            .OrderBy(p => p.GroupType, StringComparer.Ordinal)
-            .Select(p => p.GroupType + "=" + p.GroupKey);
+            .OrderBy(g => g.GroupType, StringComparer.Ordinal)
+            .Select(g => Uri.EscapeDataString(g.GroupType) + "=" + Uri.EscapeDataString(g.GroupKey));
         return string.Join(";", pairs);
     }
 
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -729,9 +729,13 @@ public sealed class PostHogClient : IPostHogClient
         GroupCollection? groups,
         CancellationToken cancellationToken)
     {
+        var groupsCacheKey = CanonicalGroupsCacheKey(groups);
+
         _featureFlagCalledEventCache.GetOrCreate(
-            key: (distinctId, featureKey, cacheKeyValue),
-            // This factory only runs when the (distinct id, key, value) tuple is not yet cached.
+            key: (distinctId, featureKey, cacheKeyValue, groupsCacheKey),
+            // This factory only runs when the (distinct id, key, value, groups) tuple is not yet cached.
+            // Group context is included so group-scoped flags fire a separate event for each group
+            // a user is evaluated under, instead of dedup-ing across groups.
             factory: cacheEntry =>
             {
                 cacheEntry.SetSize(1);
@@ -757,6 +761,23 @@ public sealed class PostHogClient : IPostHogClient
                     _options.Value.FeatureFlagSentCacheCompactionPercentage),
                 cancellationToken);
         }
+    }
+
+    // Build a canonical string for the groups collection that is order-independent
+    // so two equal sets of groups always produce the same dedup cache key.
+    // Returns the empty string when no groups were passed, which keeps the legacy
+    // "no groups" dedupe shape for callers that don't pass a GroupCollection.
+    static string CanonicalGroupsCacheKey(GroupCollection? groups)
+    {
+        if (groups is null || groups.Count == 0)
+        {
+            return string.Empty;
+        }
+        var pairs = groups
+            .Select(g => (g.GroupType, g.GroupKey))
+            .OrderBy(p => p.GroupType, StringComparer.Ordinal)
+            .Select(p => p.GroupType + "=" + p.GroupKey);
+        return string.Join(";", pairs);
     }
 
     sealed class EvaluationsHost : IFeatureFlagEvaluationsHost
--- a/tests/UnitTests/Features/FeatureFlagsTests.cs
+++ b/tests/UnitTests/Features/FeatureFlagsTests.cs
@@ -295,6 +295,82 @@ public class TheIsFeatureFlagEnabledAsyncMethod
     }
 
     [Fact]
+    public async Task CapturesSeparateFeatureFlagCalledEventsPerGroupContext()
+    {
+        var container = new TestContainer();
+        var messageHandler = container.FakeHttpMessageHandler;
+        messageHandler.AddRepeatedFlagsResponse(2, """{"featureFlags": {"flag-key": true}}""");
+        var captureRequestHandler = messageHandler.AddBatchResponse();
+        var client = container.Activate<PostHogClient>();
+
+        // Same user, same flag, same response — but two different group contexts.
+        var optionsA = new FeatureFlagOptions { Groups = new GroupCollection() };
+        optionsA.Groups!.Add(new Group("organization", "org-a"));
+        var optionsB = new FeatureFlagOptions { Groups = new GroupCollection() };
+        optionsB.Groups!.Add(new Group("organization", "org-b"));
+
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsA, CancellationToken.None));
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsB, CancellationToken.None));
+
+        await client.FlushAsync();
+        var body = captureRequestHandler.GetReceivedRequestBody(indented: false);
+        var matches = System.Text.RegularExpressions.Regex.Matches(body, "\\$feature_flag_called");
+        Assert.Equal(2, matches.Count);
+        Assert.Contains("\"organization\":\"org-a\"", body, StringComparison.Ordinal);
+        Assert.Contains("\"organization\":\"org-b\"", body, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public async Task DedupesFeatureFlagCalledAcrossRepeatedCallsUnderSameGroupContext()
+    {
+        var container = new TestContainer();
+        var messageHandler = container.FakeHttpMessageHandler;
+        messageHandler.AddRepeatedFlagsResponse(3, """{"featureFlags": {"flag-key": true}}""");
+        var captureRequestHandler = messageHandler.AddBatchResponse();
+        var client = container.Activate<PostHogClient>();
+
+        var options = new FeatureFlagOptions { Groups = new GroupCollection() };
+        options.Groups!.Add(new Group("organization", "org-a"));
+
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", options, CancellationToken.None));
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", options, CancellationToken.None));
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", options, CancellationToken.None));
+
+        await client.FlushAsync();
+        var body = captureRequestHandler.GetReceivedRequestBody(indented: false);
+        var matches = System.Text.RegularExpressions.Regex.Matches(body, "\\$feature_flag_called");
+        Assert.Single(matches);
+    }
+
+    [Fact]
+    public async Task DedupesFeatureFlagCalledAcrossGroupInsertionOrder()
+    {
+        var container = new TestContainer();
+        var messageHandler = container.FakeHttpMessageHandler;
+        messageHandler.AddRepeatedFlagsResponse(2, """{"featureFlags": {"flag-key": true}}""");
+        var captureRequestHandler = messageHandler.AddBatchResponse();
+        var client = container.Activate<PostHogClient>();
+
+        // Same groups, two different insertion orders. The dedup key is canonicalized so both
+        // calls produce a cache hit.
+        var optionsA = new FeatureFlagOptions { Groups = new GroupCollection() };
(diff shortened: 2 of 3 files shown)


src/PostHog/PostHogClient.cs
```diff
@@ -759,6 +763,23 @@ void TryCaptureDedupedFeatureFlagCalledEvent(
...
+    // so two equal sets of groups always produce the same dedup cache key.
+    // Returns the empty string when no groups were passed, which keeps the legacy
+    // "no groups" dedupe shape for callers that don't pass a GroupCollection.
+    static string CanonicalGroupsCacheKey(GroupCollection? groups)
+    {
+        if (groups is null || groups.Count == 0)
+        {
+            return string.Empty;
+        }
+        var pairs = groups
```
haacked, line 776: `suggestion`: `GroupType` and `GroupKey` are `required string` with no charset validation (see `Group.cs`), and the encoding here joins them with `=` and `;` without escaping. A `Group { GroupType = "organization", GroupKey = "a;team=b" }` produces the same canonical key as two groups `{organization=a, team=b}`. The failure mode is the exact bug this PR is fixing (a dedup collision), just narrowed to the case where a group field contains `;` or `=`. Probably out-of-contract for callers in practice, but length-prefixing closes it cleanly:

```suggestion
        var pairs = groups
            .Select(g => (g.GroupType, g.GroupKey))
            .OrderBy(p => p.GroupType, StringComparer.Ordinal)
            .Select(p => $"{p.GroupType.Length}:{p.GroupType}={p.GroupKey.Length}:{p.GroupKey}");
```

Your call on whether to fix here or document the contract on `Group`.

src/PostHog/PostHogClient.cs
```diff
@@ -759,6 +763,23 @@ void TryCaptureDedupedFeatureFlagCalledEvent(
         }
     }
 
+    // Build a canonical string for the groups collection that is order-independent
+    // so two equal sets of groups always produce the same dedup cache key.
+    // Returns the empty string when no groups were passed, which keeps the legacy
+    // "no groups" dedupe shape for callers that don't pass a GroupCollection.
+    static string CanonicalGroupsCacheKey(GroupCollection? groups)
```
haacked, line 770: `suggestion`: This is the second order-independent group-canonicalizer in the repo. `FeatureFlagCacheKey.Generate` at `src/PostHog/Features/FeatureFlagCacheKey.cs:38-65` already does the same job for the flag-evaluation cache, using `OrderBy(g => g.GroupType)` (default culture-sensitive comparer) and `,` between entries. This one uses `StringComparer.Ordinal` and `;`. Two helpers solving the same sub-problem will drift the next time either is touched, and a future reader will assume the difference is meaningful when it isn't.

Worth a follow-up (not this PR): align both on `StringComparer.Ordinal` (the right choice for cache keys — deterministic across cultures, faster), or factor a small `Group.OrderedByType(IEnumerable<Group>)` helper that both call sites compose from.

Same point on placement: this helper doesn't touch any `PostHogClient` state, so it would sit more naturally next to `GroupCollection`.

tests/UnitTests/Features/FeatureFlagsTests.cs
```diff
@@ -294,6 +294,82 @@ public async Task CapturesFeatureFlagCalledEventOnlyOncePerDistinctIdFlagKeyAndR
...
+
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsA, CancellationToken.None));
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsB, CancellationToken.None));
+
+        await client.FlushAsync();
+        var body = captureRequestHandler.GetReceivedRequestBody(indented: false);
+        var matches = System.Text.RegularExpressions.Regex.Matches(body, "\\$feature_flag_called");
+        Assert.Equal(2, matches.Count);
+        Assert.Contains("\"organization\":\"org-a\"", body, StringComparison.Ordinal);
+        Assert.Contains("\"organization\":\"org-b\"", body, StringComparison.Ordinal);
```
haacked: `suggestion`: The comment on `CanonicalGroupsCacheKey` documents that null and empty `GroupCollection` both collapse to the same dedup key, but no test pins that. If a future refactor changes the empty branch to return `"<empty>"` or similar, callers that mix null and empty start double-firing `$feature_flag_called` and nothing in CI catches it.

```csharp
[Fact]
public async Task DedupesFeatureFlagCalledAcrossNullAndEmptyGroupCollections()
{
    var messageHandler = container.FakeHttpMessageHandler;
    messageHandler.AddRepeatedFlagsResponse(2, """{"featureFlags": {"flag-key": true}}""");
    var captureRequestHandler = messageHandler.AddBatchResponse();

    var optionsNull = new FeatureFlagOptions { Groups = null };
    var optionsEmpty = new FeatureFlagOptions { Groups = new GroupCollection() };

    Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsNull, CancellationToken.None));
    Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsEmpty, CancellationToken.None));

    await client.FlushAsync();
    var body = captureRequestHandler.GetReceivedRequestBody(indented: false);
    var matches = System.Text.RegularExpressions.Regex.Matches(body, "\\$feature_flag_called");
    Assert.Single(matches);
}
```

tests/UnitTests/Features/FeatureFlagsTests.cs
```diff
@@ -294,6 +294,82 @@ public async Task CapturesFeatureFlagCalledEventOnlyOncePerDistinctIdFlagKeyAndR
...
+        var optionsB = new FeatureFlagOptions { Groups = new GroupCollection() };
+        optionsB.Groups!.Add(new Group("organization", "org-b"));
+
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsA, CancellationToken.None));
+        Assert.True(await client.IsFeatureEnabledAsync("flag-key", "user-1", optionsB, CancellationToken.None));
+
+        await client.FlushAsync();
+        var body = captureRequestHandler.GetReceivedRequestBody(indented: false);
+        var matches = System.Text.RegularExpressions.Regex.Matches(body, "\\$feature_flag_called");
+        Assert.Equal(2, matches.Count);
```
haacked: `suggestion`: `Assert.Equal(2, matches.Count)` plus two `Assert.Contains` checks verifies two events fired and both org strings appear somewhere, but it doesn't bind `org-a` to one event and `org-b` to the other. If the factory closure ever captured the wrong `groups` reference and both events ended up with `org-b`, this test would still pass: two events fire because there are two distinct cache keys, both org strings appear because `org-a` shows up in the request URL or another property.

The pre-existing `CapturesFeatureFlagCalledEventWithGroupInformation` (line 2449) shows the production payload nests groups under `$groups: { "<type>": "<key>" }`. Parsing the batch and asserting one event per `$groups` value would close the gap. Not blocking; the dedup behavior is verified, just not the wire shape.

src/PostHog/PostHogClient.cs
```diff
@@ -729,9 +729,13 @@ void TryCaptureDedupedFeatureFlagCalledEvent(
         GroupCollection? groups,
         CancellationToken cancellationToken)
     {
+        var groupsCacheKey = CanonicalGroupsCacheKey(groups);
```
haacked, line 732: `question`: The changeset note is clear about the behavior change (more events when group context varies). For users on usage-based billing or tight event-volume budgets, this could be material if they evaluate the same flag against many group contexts inside the 10-min sliding TTL. The cache is bounded at 50k entries with low-priority eviction, so steady-state is just more frequent compaction (not breakage), but it's a real uplift in event volume for some workloads. Worth one extra sentence in the changeset calling out "expect more `$feature_flag_called` events when group context varies", or is the current wording already enough signal?

src/PostHog/PostHogClient.cs
```diff
@@ -759,6 +763,23 @@ void TryCaptureDedupedFeatureFlagCalledEvent(
         }
     }
 
+    // Build a canonical string for the groups collection that is order-independent
+    // so two equal sets of groups always produce the same dedup cache key.
+    // Returns the empty string when no groups were passed, which keeps the legacy
```
haacked: `nit`: "Legacy" implies a newer shape has replaced this one. Nothing about the no-groups path is legacy; it's just the current dedup shape for callers that don't pass a `GroupCollection`.

```suggestion
    // Returns the empty string when no groups were passed, preserving the
    // no-groups dedupe shape for callers that don't pass a GroupCollection.
```

src/PostHog/PostHogClient.cs
```diff
@@ -729,9 +729,13 @@ void TryCaptureDedupedFeatureFlagCalledEvent(
         GroupCollection? groups,
         CancellationToken cancellationToken)
     {
+        var groupsCacheKey = CanonicalGroupsCacheKey(groups);
+
         _featureFlagCalledEventCache.GetOrCreate(
-            key: (distinctId, featureKey, cacheKeyValue),
-            // This factory only runs when the (distinct id, key, value) tuple is not yet cached.
+            key: (distinctId, featureKey, cacheKeyValue, groupsCacheKey),
+            // This factory only runs when the (distinct id, key, value, groups) tuple is not yet cached.
```
haacked: `nit`: The comment says "the (distinct id, key, value, groups) tuple" but the tuple in code is `(distinctId, featureKey, cacheKeyValue, groupsCacheKey)`. `cacheKeyValue` is distinct from the flag value (it's the response cache key value), so a reader has to do a small translation. Matching the comment to the variable names removes that hop.

src/PostHog/PostHogClient.cs
```diff
@@ -759,6 +763,23 @@ void TryCaptureDedupedFeatureFlagCalledEvent(
...
+    // Returns the empty string when no groups were passed, which keeps the legacy
+    // "no groups" dedupe shape for callers that don't pass a GroupCollection.
+    static string CanonicalGroupsCacheKey(GroupCollection? groups)
+    {
+        if (groups is null || groups.Count == 0)
+        {
+            return string.Empty;
+        }
+        var pairs = groups
+            .Select(g => (g.GroupType, g.GroupKey))
```
haacked: `nit`: The intermediate `.Select(g => (g.GroupType, g.GroupKey))` tuple exists only so `OrderBy` can read `p.GroupType`. You can drop it and order directly on the `Group`:

```suggestion
            .OrderBy(g => g.GroupType, StringComparer.Ordinal)
            .Select(g => g.GroupType + "=" + g.GroupKey);
```

(Adjust the next two lines accordingly: this replaces lines 777-779.) One less hop, same behavior. Not worth a re-spin on its own.

ioannisj (approved):

haacked (approved):
Nice fix. A few suggestions and nits inline; nothing blocking.

src/PostHog/PostHogClient.cs
gustavohstrassburger: Since we don't actually bill based on `$feature_flag_called` events, I don't see this as super relevant to the users. I feel like this is more of an internal problem, than something we need to tell the users. Wdyt?

src/PostHog/PostHogClient.cs
gustavohstrassburger: Will address this in a follow up PR.

src/PostHog/PostHogClient.cs
haacked: > I don't see this as super relevant to the users.

You're right. Sorry for the noise.

PR #213, opened 2026-05-26
feat: add feature flag evaluation contexts

### LINKED ISSUE #202: Evaluation Contexts
### Problem Statement

I want to use evaluation contexts with the .NET SDK but I can't seem to find a solution - maybe I've missed it.

I am reading the docs here about them https://posthog.com/docs/feature-flags/evaluation-contexts

### Solution Brainstorm

_No response_

dustinbyrne:

Adds SDK-level feature flag evaluation contexts to PostHog .NET so `/flags` requests can be scoped to the application, platform, or product area that should evaluate a flag.

This matches the existing cross-SDK behavior in PostHog JS/Node and Android/server: configured contexts are sent as top-level `evaluation_contexts`, while null or empty configuration omits the field.

Resolves https://github.com/PostHog/posthog-dotnet/issues/202

- `dotnet test tests/UnitTests/UnitTests.csproj -f net8.0 --filter "FullyQualifiedName~FeatureFlagEvaluationsTests|FullyQualifiedName~RegistrationTests"`

Also ran the full net8.0 unit suite locally; it currently has 2 unrelated existing error-tracking failures in `PostHogClientTests` outside this PR's diff.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/few-flags-dance.md
+++ b/.changeset/few-flags-dance.md
@@ -1,5 +1,6 @@
 ---
 "PostHog": minor
+"PostHog.AspNetCore": minor
 ---
 
 Add feature flag evaluation contexts via `PostHogOptions.EvaluationContexts`. `/flags` requests now send `evaluation_contexts` when configured.
--- a/.changeset/few-flags-dance.md
+++ b/.changeset/few-flags-dance.md
@@ -1,0 +1,5 @@
+---
+"PostHog": minor
+---
+
+Add feature flag evaluation contexts via `PostHogOptions.EvaluationContexts`. `/flags` requests now send `evaluation_contexts` when configured.
--- a/src/PostHog/Config/PostHogOptions.cs
+++ b/src/PostHog/Config/PostHogOptions.cs
@@ -59,6 +59,15 @@ public sealed class PostHogOptions : IOptions<PostHogOptions>
     public string? PersonalApiKey { get; set; }
 
     /// <summary>
+    /// Evaluation contexts for feature flags.
+    /// </summary>
+    /// <remarks>
+    /// When set, only feature flags with no evaluation contexts or at least one matching context will be evaluated
+    /// for this SDK instance. Use this to isolate flags by application, platform, or product area.
+    /// </remarks>
+    public IReadOnlyList<string>? EvaluationContexts { get; set; }
+
+    /// <summary>
     /// Whether this client is disabled and should no-op instead of sending data to PostHog. (Default: false)
     /// </summary>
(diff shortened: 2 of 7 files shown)


tests/UnitTests/Features/FeatureFlagEvaluationsTests.cs
```diff
@@ -79,6 +79,46 @@ await client.EvaluateFlagsAsync(
         Assert.Equal(new[] { "flag-a", "flag-b" }, flagKeys);
     }
 
+    public static IEnumerable<object?[]> EvaluationContextsRequestBodyCases()
+    {
+        yield return [new[] { "production", "backend" }, new[] { "production", "backend" }];
```
haacked: `suggestion`: The positive case has two elements, which catches at most one wrong order. A single-element case would confirm the JSON shape stays an array (not a scalar) for `Count == 1`, and a three-element case actually has enough permutations to detect a reorder regression:

```csharp
yield return [new[] { "single-context" }, new[] { "single-context" }];
yield return [new[] { "production", "backend", "internal" }, new[] { "production", "backend", "internal" }];
```

src/PostHog/Config/PostHogOptions.cs
```diff
@@ -58,6 +58,15 @@ public string? ProjectApiKey
...
     public string? PersonalApiKey { get; set; }
 
+    /// <summary>
+    /// Evaluation contexts for feature flags.
+    /// </summary>
+    /// <remarks>
+    /// When set, only feature flags with no evaluation contexts or at least one matching context will be evaluated
+    /// for this SDK instance. Use this to isolate flags by application, platform, or product area.
+    /// </remarks>
+    public IReadOnlyList<string>? EvaluationContexts { get; set; }
```
haacked: `nit`: Other feature-flag-specific options on this class use the `FeatureFlag` prefix (`FeatureFlagPollInterval`, `FeatureFlagSentCacheSizeLimit`). `EvaluationContexts` reads ambiguously next to general-purpose options like `Disabled` and `HostUrl`. `FeatureFlagEvaluationContexts` would match the convention; trade-off is divergence from posthog-js's `evaluation_contexts` wire-name. Your call.

.changeset/few-flags-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"PostHog": minor
```
marandaneto, line 2: you also need the asp net core package here

marandaneto (approved):

PR #219, opened 2026-06-02
feat: emit $is_server property on captured events

turnipdabeets:
Adds `$is_server: true` to every event captured by this SDK so PostHog ingestion can identify server-side events.

### DIFF
--- a/src/PostHog/Api/CapturedEvent.cs
+++ b/src/PostHog/Api/CapturedEvent.cs
@@ -16,11 +16,17 @@ public class CapturedEvent
     /// <param name="distinctId">The identifier for the user.</param>
     /// <param name="properties">The properties to associate with the event.</param>
     /// <param name="timestamp">The ISO 8601 timestamp.</param>
+    /// <param name="isServer">
+    /// When <c>true</c> (the default), the event includes the <c>$is_server</c> property set to <c>true</c>. Set to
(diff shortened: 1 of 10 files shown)


src/PostHog/Api/CapturedEvent.cs
```diff
@@ -16,11 +16,17 @@ public class CapturedEvent
...
+    /// <c>false</c> (via <see cref="PostHogOptions.IsServer"/>) when using the SDK as a client/CLI so the device OS
+    /// is attributed normally; the property is then omitted.
+    /// </param>
     public CapturedEvent(
         string eventName,
         string distinctId,
         Dictionary<string, object>? properties,
-        DateTimeOffset timestamp)
+        DateTimeOffset timestamp,
+        bool isServer = true)
```
dustinbyrne: this looks like a public API - do we need this addition?

We could write `$is_server` in `PostHogClient.CaptureCore` after creating the captured event. That'd also allow us to avoid changing `ExceptionPropertiesBuilder`, as exception events would also flow through `CaptureCore`.

I think it's worth it because it limits the public API scope creep and keeps `config.IsServer` as the single authority

turnipdabeets: updated

dustinbyrne (approved):

PR #221, opened 2026-06-04
feat(flags): support early_exit in local evaluation

gustavohstrassburger:

PostHog added an "early exit" option to feature flags in the server-side (Rust) evaluation engine, and this has already been ported to posthog-node and posthog-python. This PR ports the same behavior to the .NET SDK's local feature flag evaluation.

When a flag's `filters.early_exit` is `true`, local condition evaluation must stop and return a definitive disabled result as soon as a condition group's property filters match (or the group has no property filters) but the rollout percentage excludes the user — instead of falling through to evaluate later condition groups.

The three per-group outcomes are now modeled explicitly, mirroring the Rust engine:

- **MATCH** — property filters matched (or none) AND rollout included the user → flag matches (existing behavior).
- **NO_MATCH** — a property filter did not match → always continue to the next group, even when `early_exit` is enabled.
- **OUT_OF_ROLLOUT_BOUND** — property filters matched (or none) but the rollout excluded the user → with `early_exit` enabled, return a definitive disabled result immediately.

When `early_exit` is absent or `false`, existing behavior is preserved exactly.

### Changes

- Added the `early_exit` field (`EarlyExit`, JSON `early_exit`, default `false`) to `FeatureFlagFilters`, including equality/hash-code.
- Converted the single-group matcher `IsConditionMatch` (bool) into `EvaluateConditionMatch` returning a tri-state `ConditionMatchResult` enum, distinguishing rollout-exclusion (`hash > rollout/100`) from property-mismatch.
- Short-circuit to `false` in the condition-group loop when `early_exit` is on and the outcome is `OutOfRolloutBound`.

Added unit tests in `TheEarlyExitBehavior` covering:

1. `early_exit` enabled → `false` without evaluating a later matching group (first group: matching props + rollout 0; second group: matching props + rollout 100).
2. `early_exit` unset (absent in JSON) → falls through to the matching second group (`true`) — regression.
3. `early_exit` explicitly `false` → falls through (`true`).
4. Property-filter failure (not rollout) does NOT early-exit even when `early_exit` is enabled → falls through (`true`).

All 889 existing unit tests still pass (`dotnet test`, net8.0). Release build is clean across all target frameworks with warnings-as-errors, and `dotnet format --verify-no-changes` reports no formatting issues.

## :pencil: Checklist

### If releasing new changes

---

### DIFF
--- a/src/PostHog/Features/LocalEvaluator.cs
+++ b/src/PostHog/Features/LocalEvaluator.cs
@@ -269,66 +269,38 @@ internal sealed class LocalEvaluator
 
         foreach (var condition in flagConditions)
         {
-            try
+            // Per-condition aggregation overrides only when the condition explicitly
+            // sets its own AggregationGroupTypeIndex (mixed targeting). When absent,
+            // fall back to the flag-level aggregation so existing pure person and
+            // pure group flags keep their original behavior.
+            var conditionAggregation = condition.AggregationGroupTypeIndex ?? flagAggregation;
+
+            var effectiveProperties = properties;
+            var effectiveBucketingId = distinctId;
+
+            // The condition explicitly sets its own aggregation, different from the
+            // flag level. Re-resolve properties and bucketing id from the condition's
+            // group so this condition evaluates against that group.
+            // conditionAggregation is non-null here: if it were null and flagAggregation
+            // were also null they'd be equal, and the only way conditionAggregation can
+            // be null at all (after the ?? above) is that case.
+            if (conditionAggregation != flagAggregation)
             {
-                // Per-condition aggregation overrides only when the condition explicitly
-                // sets its own AggregationGroupTypeIndex (mixed targeting). When absent,
-                // fall back to the flag-level aggregation so existing pure person and
-                // pure group flags keep their original behavior.
-                var conditionAggregation = condition.AggregationGroupTypeIndex ?? flagAggregation;
-
-                var effectiveProperties = properties;
-                var effectiveBucketingId = distinctId;
-
-                // The condition explicitly sets its own aggregation, different from the
-                // flag level. Re-resolve properties and bucketing id from the condition's
-                // group so this condition evaluates against that group.
-                // conditionAggregation is non-null here: if it were null and flagAggregation
-                // were also null they'd be equal, and the only way conditionAggregation can
-                // be null at all (after the ?? above) is that case.
-                if (conditionAggregation != flagAggregation)
+                if (!_groupTypeMapping.TryGetValue(conditionAggregation!.Value, out var groupType)
+                    || groups is null
+                    || !groups.TryGetGroup(groupType, out var group))
                 {
-                    if (!_groupTypeMapping.TryGetValue(conditionAggregation!.Value, out var groupType)
-                        || groups is null
-                        || !groups.TryGetGroup(groupType, out var group))
-                    {
-                        // Skip this condition: group type unknown or not passed in.
-                        continue;
-                    }
-                    effectiveProperties = group.Properties;
-                    effectiveBucketingId = group.GroupKey;
-                }
-
-                var conditionResult = EvaluateConditionMatch(flag, effectiveBucketingId, condition, effectiveProperties, evaluationCache, groups);
-
-                // When early_exit is enabled and the condition's property filters matched (or
-                // there were none) but the rollout excluded the user, stop evaluating later
-                // condition groups and return a definitive disabled result. Mirrors the
-                // server-side Rust evaluation engine. A pure property-filter mismatch always
-                // falls through to the next condition group, even when early_exit is enabled.
-                if (earlyExit && conditionResult == ConditionMatchResult.OutOfRolloutBound)
-                {
-                    if (isInconclusive)
-                    {
-                        throw new InconclusiveMatchException("Can't determine if feature flag is enabled or not with given properties");
-                    }
-                    return false;
-                }
-
-                if (conditionResult != ConditionMatchResult.Match)
-                {
+                    // Skip this condition: group type unknown or not passed in.
                     continue;
                 }
+                effectiveProperties = group.Properties;
+                effectiveBucketingId = group.GroupKey;
+            }
 
-                var variantOverride = condition.Variant;
-                var variant = variantOverride is not null
-                              && flagVariants.Select(v => v.Key).Contains(variantOverride)
-                    ? variantOverride
-                    : GetMatchingVariant(flag, effectiveBucketingId);
-
-                return variant is not null
-                    ? new StringOrValue<bool>(variant)
-                    : true;
+            ConditionMatchResult? conditionResult = null;
+            try
+            {
+                conditionResult = EvaluateConditionMatch(flag, effectiveBucketingId, condition, effectiveProperties, evaluationCache, groups);
             }
             catch (RequiresServerEvaluationException)
             {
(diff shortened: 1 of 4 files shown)


gustavohstrassburger:
@copilot fix the errors in https://github.com/PostHog/posthog-dotnet/actions/runs/27026012051/job/79766113172?pr=221

Error: /home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/Features/LocalEvaluatorTests.cs(2708,23): error WHITESPACE: Fix whitespace formatting. Delete 1 characters. [/home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/UnitTests.csproj]
Error: /home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/Features/LocalEvaluatorTests.cs(2708,46): error WHITESPACE: Fix whitespace formatting. Delete 1 characters. [/home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/UnitTests.csproj]
Error: /home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/Features/LocalEvaluatorTests.cs(2709,46): error WHITESPACE: Fix whitespace formatting. Delete 1 characters. [/home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/UnitTests.csproj]
Error: /home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/Features/LocalEvaluatorTests.cs(2710,23): error WHITESPACE: Fix whitespace formatting. Delete 1 characters. [/home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/UnitTests.csproj]
Error: /home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/Features/LocalEvaluatorTests.cs(2710,46): error WHITESPACE: Fix whitespace formatting. Delete 1 characters. [/home/runner/work/posthog-dotnet/posthog-dotnet/tests/UnitTests/UnitTests.csproj]

src/PostHog/Features/LocalEvaluator.cs
```diff
@@ -298,9 +299,19 @@ StringOrValue<bool> MatchFeatureFlagProperties(
...
-                // the matching variant
-                if (!IsConditionMatch(flag, effectiveBucketingId, condition, effectiveProperties, evaluationCache, groups))
+                var conditionResult = EvaluateConditionMatch(flag, effectiveBucketingId, condition, effectiveProperties, evaluationCache, groups);
+
+                // When early_exit is enabled and the condition's property filters matched (or
+                // there were none) but the rollout excluded the user, stop evaluating later
+                // condition groups and return a definitive disabled result. Mirrors the
+                // server-side Rust evaluation engine. A pure property-filter mismatch always
+                // falls through to the next condition group, even when early_exit is enabled.
+                if (earlyExit && conditionResult == ConditionMatchResult.OutOfRolloutBound)
```
dustinbyrne: Blocking correctness issue: this can mask a prior inconclusive condition.

If an earlier condition caught `InconclusiveMatchException`, `isInconclusive` is set, but this early-exit branch returns `false` before the post-loop fallback at lines 342-344 can throw. That makes local evaluation return a definitive disabled result when it should fall back to server evaluation.

Repro shape:
1. earlier condition is inconclusive locally, e.g. missing required property
2. later condition evaluates to `OutOfRolloutBound`
3. `early_exit=true`

Current result: `false`
Expected: throw `InconclusiveMatchException` / server fallback

Please check `isInconclusive` before returning `false` here; if it’s true, preserve fallback semantics instead of short-circuiting.

dustinbyrne (commented):
Validated the early_exit edge case locally; leaving one blocking correctness note inline.

tests/UnitTests/Features/LocalEvaluatorTests.cs
```diff
@@ -2600,4 +2600,166 @@ public void DoesNotMatchTypeMismatches(object expected, object actual, bool shou
...
+            [
+                new LocalFeatureFlag
+                {
+                    Id = 42,
+                    TeamId = 23,
+                    Name = "early-exit-feature-flag",
+                    Key = "early-exit",
+                    Filters = new FeatureFlagFilters
+                    {
+                        EarlyExit = earlyExit,
```
haacked, line 2628: `suggestion`: No test ever deserializes `early_exit: true` from JSON. Every positive early-exit case sets `EarlyExit = earlyExit` through this object initializer, which never touches the [`[JsonPropertyName("early_exit")]`](https://github.com/posthog/posthog-dotnet/blob/b5b0488675fec02ca7eb6dcaf3a89dacc4d6da51/src/PostHog/Api/LocalEvaluationApiResult.cs#L154) binding. `CreateFlagsWithoutEarlyExitField` only proves the absent field defaults to `false`. So a typo in that property name would pass every test here while real payloads carrying `"early_exit": true` would deserialize to `false` and quietly disable the feature in production. Worth one JSON round-trip that injects `"early_exit": true` and asserts the out-of-rollout group early-exits.

dustinbyrne (commented):
Take a look at the following test, it's still a failing condition. Once the early exit / out of rollout band condition is reached, later groups shouldn't be considered. But because an earlier condition was inconclusive, local evaluation also can’t safely return false, so it should preserve the inconclusive/server-fallback path.

```csharp
[Fact]
public void ThrowsInconclusiveWhenEarlyExitOutOfRolloutHasLaterMatchingGroup()
{
    var flags = new LocalEvaluationApiResult
    {
        Flags =
        [
            new LocalFeatureFlag
            {
                Id = 42,
                TeamId = 23,
                Name = "early-exit-feature-flag",
                Key = "early-exit",
                Filters = new FeatureFlagFilters
                {
                    EarlyExit = true,
                    Groups =
                    [
                        new FeatureFlagGroup
                        {
                            Properties =
                            [
                                new PropertyFilter
                                {
                                    Type = FilterType.Person,
                                    Key = "missing_property",
                                    Value = new PropertyFilterValue("some_value"),
                                    Operator = ComparisonOperator.Exact
                                }
                            ],
                            RolloutPercentage = 100
                        },
                        new FeatureFlagGroup
                        {
                            Properties =
                            [
                                new PropertyFilter
                                {
                                    Type = FilterType.Person,
                                    Key = "email",
                                    Value = new PropertyFilterValue("tyrion@example.com"),
                                    Operator = ComparisonOperator.Exact
                                }
                            ],
                            RolloutPercentage = 0
                        },
                        new FeatureFlagGroup
                        {
                            Properties =
                            [
                                new PropertyFilter
                                {
                                    Type = FilterType.Person,
                                    Key = "email",
                                    Value = new PropertyFilterValue("tyrion@example.com"),
                                    Operator = ComparisonOperator.Exact
                                }
                            ],
                            RolloutPercentage = 100
                        }
                    ]
                }
            }
        ],
        GroupTypeMapping = new Dictionary<string, string>()
    };

    var localEvaluator = new LocalEvaluator(flags);

    Assert.Throws<InconclusiveMatchException>(() => localEvaluator.EvaluateFeatureFlag(
        key: "early-exit",
        distinctId: "1234",
        personProperties: new Dictionary<string, object?> { ["email"] = "tyrion@example.com" }));
}
```

haacked (commented):
Agree with @dustinbyrne's suggestions. Also, one test suggestion.

src/PostHog/Features/LocalEvaluator.cs
```diff
@@ -338,7 +338,14 @@ StringOrValue<bool> MatchFeatureFlagProperties(
             catch (InconclusiveMatchException)
             {
                 // Evaluation error (bad regex, invalid date, missing property, etc.)
-                // Track that we had an inconclusive match, but try other conditions
+                // If early_exit is enabled and a prior condition was already inconclusive,
+                // we cannot safely skip to later groups — re-throw so the caller falls back
+                // to the server, which mirrors what would have happened had the rollout
+                // boundary short-circuited first.
+                if (earlyExit && isInconclusive)
```
dustinbyrne: `earlyExit` reflects the flag level setting, yes? It's not indicating that for this specific evaluation, we've missed the rollout band, and we should _now_ exit early.

That makes this too broad, and the following test case would now fail:

Given a flag with `EarlyExit: true` and the following condition groups:
1. inconclusive group with rollout 100
2. another inconclusive group with rollout 100
3. matching group with rollout 100

This case should still return `true`, but it'll now throw `InconclusiveMatchException` before reaching the last matching group

gustavohstrassburger: yeah, you're right, I'll fix this, thanks dustin.

dustinbyrne (approved):
nice!

PR #222, opened 2026-06-04
refactor: reduce dry4dotnet duplication

marandaneto:

Refactor duplicate code paths found by the dry4dotnet PoC (github.com/marandaneto/dry4dotnet) without changing the public API or behavior.

This reduces repeated implementation/test patterns in group identify helpers, no-op client methods, capture helpers, sample setup, fake HTTP response helpers, and several duplicate-heavy tests.

- `dotnet build PostHog.sln --no-restore`
- `dotnet test PostHog.sln --no-build --no-restore`
- `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --no-restore --filter "FullyQualifiedName~SemanticVersionTests"`
- `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --no-restore --filter "FullyQualifiedName~TheIdentifyGroupAsyncMethod|FullyQualifiedName~TheCapturePageViewMethod|FullyQualifiedName~TheNoOpPostHogClient"`
- `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --no-build --no-restore --filter "FullyQualifiedName~SemanticVersionTests|FullyQualifiedName~LocalEvaluatorTests|FullyQualifiedName~FeatureFlagErrorTracking"`
- Re-ran dry4dotnet PoC and reduced candidates from 284 to 126.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/src/PostHog/NoOpPostHogClient.cs
+++ b/src/PostHog/NoOpPostHogClient.cs
@@ -32,7 +32,7 @@ internal sealed class NoOpPostHogClient : IPostHogClient, IFeatureFlagEvaluation
         StringOrValue<int> key,
         Dictionary<string, object>? properties,
         CancellationToken cancellationToken)
-        => GroupIdentifyAsync(string.Empty, type, key, properties, cancellationToken);
+        => NoOpApiResultTask;
 
     public Task<ApiResult> GroupIdentifyAsync(
         string distinctId,
@@ -76,7 +76,7 @@ internal sealed class NoOpPostHogClient : IPostHogClient, IFeatureFlagEvaluation
         GroupCollection? groups,
         FeatureFlagEvaluations? flags,
         DateTimeOffset? timestamp = null)
-        => Capture(distinctId, exception?.GetType().FullName ?? string.Empty, properties, groups, flags, timestamp);
+        => false;
 
     public Task<bool> IsFeatureEnabledAsync(
         string featureKey,
--- a/src/PostHog/NoOpPostHogClient.cs
+++ b/src/PostHog/NoOpPostHogClient.cs
@@ -9,6 +9,7 @@ namespace PostHog.Sdk;
 internal sealed class NoOpPostHogClient : IPostHogClient, IFeatureFlagEvaluationsHost
 {
     static readonly IReadOnlyDictionary<string, FeatureFlag> EmptyFeatureFlags = new Dictionary<string, FeatureFlag>();
+    static readonly Task<ApiResult> NoOpApiResultTask = Task.FromResult(new ApiResult(0));
 
     NoOpPostHogClient()
     {
@@ -17,21 +18,21 @@ internal sealed class NoOpPostHogClient : IPostHogClient, IFeatureFlagEvaluation
     internal static NoOpPostHogClient Instance { get; } = new();
 
     public Task<ApiResult> AliasAsync(string previousId, string newId, CancellationToken cancellationToken)
-        => Task.FromResult(new ApiResult(0));
+        => NoOpApiResultTask;
 
     public Task<ApiResult> IdentifyAsync(
         string distinctId,
         Dictionary<string, object>? personPropertiesToSet,
         Dictionary<string, object>? personPropertiesToSetOnce,
         CancellationToken cancellationToken)
-        => Task.FromResult(new ApiResult(0));
+        => NoOpApiResultTask;
 
     public Task<ApiResult> GroupIdentifyAsync(
         string type,
         StringOrValue<int> key,
         Dictionary<string, object>? properties,
         CancellationToken cancellationToken)
-        => Task.FromResult(new ApiResult(0));
+        => GroupIdentifyAsync(string.Empty, type, key, properties, cancellationToken);
 
     public Task<ApiResult> GroupIdentifyAsync(
         string distinctId,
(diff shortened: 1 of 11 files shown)


marandaneto:
Addressed Greptile feedback in 0c6c389: restored escaped tab test data for readability and made the test dispatch fail loudly for unknown range/operator strings. Verified with `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --no-restore --filter "FullyQualifiedName~SemanticVersionTests"`.

dustinbyrne (approved):

src/PostHog/NoOpPostHogClient.cs
```diff
@@ -75,7 +76,7 @@ public bool CaptureException(
         GroupCollection? groups,
         FeatureFlagEvaluations? flags,
         DateTimeOffset? timestamp = null)
-        => false;
+        => Capture(distinctId, exception?.GetType().FullName ?? string.Empty, properties, groups, flags, timestamp);
```
haacked: `suggestion`: This no-op `CaptureException` computes `exception?.GetType().FullName` only to hand it to `Capture`, which always returns `false` and throws the value away. It's wasted work in a no-op, and the null-conditional makes a reader pause over a name that never gets used. The other no-op leaves stop at the result, so returning `false` directly is simpler and more consistent:

```suggestion
        => false;
```

tests/UnitTests/Library/SemanticVersionTests.cs
```diff
@@ -322,75 +152,60 @@ public void CalculatesCorrectCaretBoundsForMajorAndMinorZero(string input, strin
         Assert.NotNull(expectedLowerVersion);
         Assert.NotNull(expectedUpperVersion);
 
-        var (lower, upper) = version.Value.GetCaretBounds();
+        var (lower, upper) = GetBounds(kind, version.Value);
```
haacked: `suggestion`: `kind` here is a stringly-typed discriminator ("tilde"/"caret") threaded through every `InlineData` and switched on with an `ArgumentOutOfRangeException` default. Unlike the semver-operator merge that kept the real `ComparisonOperator` enum, a typo like `[InlineData("Tilde", …)]` compiles fine and only fails at runtime in the default arm. An enum discriminator (`enum BoundKind { Tilde, Caret }`) makes the switch exhaustive and the rows impossible to mistype. Your call, the strings work today.

tests/UnitTests/Library/SemanticVersionTests.cs
```diff
@@ -171,148 +99,50 @@ public class TheCompareToMethod
...
 
         Assert.NotNull(leftVersion);
         Assert.NotNull(rightVersion);
-        Assert.Equal(expected, Math.Sign(leftVersion.Value.CompareTo(rightVersion.Value)));
+        Assert.Equal(expectedSign, Math.Sign(leftVersion.Value.CompareTo(rightVersion.Value)));
     }
 }
 
-public class TheGetTildeBoundsMethod
+public class TheVersionRangeBoundsMethods
```
haacked: `question`: This merges the old `TheGetTildeBoundsMethod` and `TheGetCaretBoundsMethod` into one class covering two distinct production methods, so the `The<Method>` naming convention used elsewhere in this file (`TheTryParseMethod`, `TheCompareToMethod`, …) no longer holds. Was collapsing them intentional? Keeping two classes, each with its own merged theory, would preserve both the DRY win and the convention.

src/PostHog/NoOpPostHogClient.cs
```diff
@@ -17,29 +18,29 @@ internal sealed class NoOpPostHogClient : IPostHogClient, IFeatureFlagEvaluation
...
-        => Task.FromResult(new ApiResult(0));
+        => NoOpApiResultTask;
 
     public Task<ApiResult> GroupIdentifyAsync(
         string type,
         StringOrValue<int> key,
         Dictionary<string, object>? properties,
         CancellationToken cancellationToken)
-        => Task.FromResult(new ApiResult(0));
+        => GroupIdentifyAsync(string.Empty, type, key, properties, cancellationToken);
```
haacked: `nit`: Routing the no-distinctId overload through the distinctId one with `string.Empty` reads as if an empty distinct id were meaningful here. It's harmless since the target is a no-op, but the sibling no-ops in this same change point at `NoOpApiResultTask` directly. Doing the same here would be more consistent and avoid implying `string.Empty` carries semantics.

haacked (approved):
Nice! A few non-blocking suggestions and nits.

src/PostHog/NoOpPostHogClient.cs
marandaneto: Good catch! Fixed in 8f1878e — the no-op CaptureException overload now returns false directly instead of routing through Capture.

src/PostHog/NoOpPostHogClient.cs
marandaneto: Good point. Fixed in 8f1878e — the no-distinctId no-op GroupIdentifyAsync overload now returns the cached NoOpApiResultTask directly.

tests/UnitTests/Library/SemanticVersionTests.cs
marandaneto: Agreed. Fixed in 8f1878e — the bounds tests no longer thread a string discriminator through InlineData; the tilde/caret cases are split by production method.

tests/UnitTests/Library/SemanticVersionTests.cs
marandaneto: Yes, preserving the The<Method>Method convention is cleaner here. Fixed in 8f1878e — split these back into TheGetTildeBoundsMethod and TheGetCaretBoundsMethod while keeping shared assertion helpers.

PR #230, opened 2026-06-12
chore: add public API baseline checks

marandaneto:
## :bulb: Motivation and Context

Add PublicApiAnalyzers baseline checks so intentional public API changes are tracked explicitly across the SDK packages.

## :green_heart: How did you test it?

- `dotnet build --configuration Release --no-restore --nologo`

## :pencil: Checklist

- [x] I reviewed the submitted code.
- [ ] I added tests to verify the changes.
- [x] I updated the docs if needed.
- [x] No breaking change or entry added to the changelog.

### If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file

PR #231, opened 2026-06-15
test: Add public API baseline probe

marandaneto:

Adds a tiny test-only public API surface so the PublicAPI baseline check added in #230 can be observed failing when the baseline is not updated.

- Ran `dotnet build --configuration Release --nologo` and confirmed it fails with `RS0016` because `PublicAPI.Shipped.txt` / `PublicAPI.Unshipped.txt` were intentionally not updated.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/src/PostHog/Examples.cs
+++ b/src/PostHog/Examples.cs
@@ -9,6 +9,11 @@ namespace PostHog;
 /// </summary>
 public static class Examples
 {
+    /// <summary>
+    /// Test-only public API marker used to verify baseline checks.
+    /// </summary>
+    public const string PublicApiBaselineProbe = "public-api-baseline-probe";
+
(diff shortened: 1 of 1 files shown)


marandaneto:
`Error: /home/runner/work/posthog-dotnet/posthog-dotnet/src/PostHog/Examples.cs(15,25): error RS0016: Symbol 'const PostHog.Examples.PublicApiBaselineProbe = "public-api-baseline-probe" -> string!' is not part of the declared public API (https://github.com/dotnet/roslyn/blob/main/src/RoslynAnalyzers/PublicApiAnalyzers/PublicApiAnalyzers.Help.md) [/home/runner/work/posthog-dotnet/posthog-dotnet/src/PostHog/PostHog.csproj::TargetFramework=net8.0]`

PR #233, opened 2026-06-15
chore: restore release checkout fetch depth

marandaneto:

Rollback the `fetch-depth: 0` removal from PR #232 for the release workflow checkout steps that need full history.

Not run; workflow-only rollback.

## :pencil: Checklist

### If releasing new changes


ioannisj (approved):
Just curious, why do we need this for release? I've seen this only for gitversion before, since dotnet could it be related?

PR #237, opened 2026-06-19
fix: generate SDK UUIDs as v7

marandaneto:

SDK-generated IDs should be UUID v7 so event UUIDs, personless distinct IDs, and default AI trace IDs are time-ordered instead of random UUID v4 values.

This adds an internal UUID v7 helper that uses `Guid.CreateVersion7()` when the runtime provides it, with a fallback generator for older runtimes/target frameworks. The helper is used for runtime SDK-generated IDs in `PostHog` and `PostHog.AI`.

- `dotnet test tests/UnitTests/UnitTests.csproj --filter "FullyQualifiedName~CapturedEventTests|FullyQualifiedName~PostHogContextTests"`
- `dotnet test tests/PostHog.AI.Tests/PostHog.AI.Tests.csproj --filter "FullyQualifiedName~PostHogOpenAIHandlerTests.SendAsyncCapturesEventOnSuccess"`
- `dotnet build src/PostHog/PostHog.csproj --no-restore`
- `dotnet build src/PostHog.AI/PostHog.AI.csproj --no-restore`
- `dotnet test PostHog.sln`
- `dotnet test`
- `dotnet format --verify-no-changes`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented by the `worker` agent using repository-local inspection and .NET CLI validation. Follow-up review feedback added runtime use of `Guid.CreateVersion7()` when available plus deterministic fallback monotonicity coverage.

### DIFF
--- a/src/PostHog/Library/UuidV7.cs
+++ b/src/PostHog/Library/UuidV7.cs
@@ -1,3 +1,4 @@
+using System.Reflection;
 using System.Security.Cryptography;
 
 namespace PostHog.Library;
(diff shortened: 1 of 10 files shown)


src/PostHog/Library/UuidV7.cs
```diff
@@ -0,0 +1,79 @@
+using System.Security.Cryptography;
+
+namespace PostHog.Library;
+
+internal static class UuidV7
```
marandaneto, line 6: does dotnet does not have a built in v7 in newer versions?
can we use the v7 if its available and fallback to this one if its running an older version?

src/PostHog/Library/UuidV7.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        for (var i = LastRandom.Length - 1; i >= 0; i--)
+        {
+            LastRandom[i]++;
+            if (LastRandom[i] != 0)
+            {
+                return;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
marandaneto: Good catch — fixed in aa9fceb by adding a deterministic same-millisecond fallback UUIDv7 monotonicity test.

src/PostHog/Library/UuidV7.cs
marandaneto: Yep — fixed in aa9fceb. `UuidV7.NewString()` now uses `Guid.CreateVersion7()` via a cached runtime reflection delegate when available, and falls back to the internal generator on older runtimes/target frameworks.

PR #243, opened 2026-06-22
Revert: "ci: set release environment for publish job"

marandaneto:
Reverts PostHog/posthog-dotnet#241

removed the requirement from nuget so we dont have to approve twice
i'll refactor this soon so we dont have to approve multiple times and still behind the environment protection


marandaneto:
version-bump already has the protection

ioannisj (approved):

PR #247, opened 2026-06-26
fix: add feature flag retry circuit breaker

marandaneto:

Stacked on top of #245 to prototype a circuit breaker for `/flags` retry behavior in a multithreaded/stateless SDK.

The existing retry is bounded per call, but during a broad connectivity issue many concurrent request threads can each sleep and retry independently. This adds a shared circuit breaker so repeated transient `/flags` transport failures put the SDK into a short cooldown and subsequent calls fail fast instead of all sleeping/retrying.

- `dotnet test tests/UnitTests/UnitTests.csproj --filter FullyQualifiedName~HttpClientExtensionsTests`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented as a draft stacked PR for review. The circuit opens after consecutive transient `/flags` failures, fails fast during cooldown, and allows a half-open probe after cooldown. HTTP/API status failures remain non-retryable and do not trip the transient-failure circuit.

### DIFF
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -1,0 +1,5 @@
+---
+'PostHog': patch
+---
+
+Retry feature flag requests after network errors only.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -1,0 +1,5 @@
+---
+'PostHog': patch
+---
+
+Retry feature flag requests after transient network errors only. The feature flag request retry count defaults to 1 and can be set to 0 to disable retries.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -1,5 +1,0 @@
----
-'PostHog': patch
----
-
-Retry feature flag requests after transient network errors only. The feature flag request retry count defaults to 1 and can be set to 0 to disable retries.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -2,4 +2,4 @@
 'PostHog': patch
 ---
 
-Retry feature flag requests after transient network errors only. The feature flag request retry count defaults to 1 and can be set to 0 to disable retries.
+Add a per-client circuit breaker for feature flag requests after consecutive transient network failures, temporarily failing fast before probing for recovery.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -2,4 +2,4 @@
 'PostHog': patch
 ---
 
-Retry feature flag requests after network errors only.
+Retry feature flag requests after transient network errors only.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -1,5 +1,6 @@
 ---
 'PostHog': patch
+'PostHog.AspNetCore': patch
 ---
 
 Add a per-client circuit breaker for feature flag requests after consecutive transient network failures, temporarily failing fast before probing for recovery.
--- a/.changeset/quiet-flags-retry.md
+++ b/.changeset/quiet-flags-retry.md
@@ -2,4 +2,4 @@
 'PostHog': patch
 ---
 
-Retry feature flag requests after transient network errors only.
+Retry feature flag requests after transient network errors only. The feature flag request retry count defaults to 1 and can be set to 0 to disable retries.
--- a/src/PostHog/Library/HttpClientExtensions.cs
+++ b/src/PostHog/Library/HttpClientExtensions.cs
@@ -14,7 +14,6 @@ namespace PostHog.Library;
 internal static class HttpClientExtensions
 {
     static readonly TimeSpan FeatureFlagInitialRetryDelay = TimeSpan.FromMilliseconds(300);
-    static readonly FeatureFlagRequestCircuitBreaker FeatureFlagCircuitBreaker = new();
 
     /// <summary>
     /// Sends a POST request to the specified Uri containing the value serialized as JSON in the request body.
@@ -56,16 +55,24 @@ internal static class HttpClientExtensions
         object content,
         TimeProvider timeProvider,
         PostHogOptions options,
-        CancellationToken cancellationToken)
+        CancellationToken cancellationToken,
+        FeatureFlagRequestCircuitBreaker? circuitBreaker = null)
     {
         var maxRetries = options.FeatureFlagRequestMaxRetries;
         var currentDelay = FeatureFlagInitialRetryDelay;
         var maxDelay = options.MaxRetryDelay;
         var attempt = 0;
+        circuitBreaker ??= new FeatureFlagRequestCircuitBreaker();
 
-        if (!FeatureFlagCircuitBreaker.TryEnter(timeProvider, out var isHalfOpenProbe))
+        if (!circuitBreaker.TryEnter(timeProvider, out var isHalfOpenProbe))
         {
             throw new HttpRequestException("Feature flag request circuit breaker is open.");
+        }
+
+        async Task DelayBeforeRetry()
+        {
+            await Delay(timeProvider, currentDelay > maxDelay ? maxDelay : currentDelay, cancellationToken);
+            currentDelay = DoubleWithCap(currentDelay, maxDelay);
         }
 
         while (true)
@@ -83,36 +90,42 @@ internal static class HttpClientExtensions
             }
             catch (HttpRequestException e) when (IsRetryableFlagsHttpRequestException(e))
             {
-                if (attempt > maxRetries ||
-                    !FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe))
+                var circuitClosed = circuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe);
+                if (attempt > maxRetries || !circuitClosed)
                 {
                     throw;
                 }
 
-                await Delay(timeProvider, currentDelay > maxDelay ? maxDelay : currentDelay, cancellationToken);
-                currentDelay = DoubleWithCap(currentDelay, maxDelay);
+                await DelayBeforeRetry();
                 continue;
             }
             catch (HttpRequestException)
             {
-                FeatureFlagCircuitBreaker.RecordNonTransientFailure(isHalfOpenProbe);
+                if (isHalfOpenProbe)
+                {
+                    circuitBreaker.RecordResponseReceived();
+                }
                 throw;
             }
             catch (TaskCanceledException) when (!cancellationToken.IsCancellationRequested)
             {
-                if (attempt > maxRetries ||
-                    !FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe))
+                var circuitClosed = circuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe);
+                if (attempt > maxRetries || !circuitClosed)
                 {
                     throw;
                 }
 
-                await Delay(timeProvider, currentDelay > maxDelay ? maxDelay : currentDelay, cancellationToken);
-                currentDelay = DoubleWithCap(currentDelay, maxDelay);
+                await DelayBeforeRetry();
                 continue;
+            }
+            catch (OperationCanceledException) when (isHalfOpenProbe && cancellationToken.IsCancellationRequested)
+            {
+                circuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe: true);
+                throw;
             }
             catch (Exception) when (isHalfOpenProbe)
             {
-                FeatureFlagCircuitBreaker.RecordNonTransientFailure(isHalfOpenProbe);
+                circuitBreaker.RecordResponseReceived();
                 throw;
             }
 
@@ -121,7 +134,10 @@ internal static class HttpClientExtensions
             // be caught by the retry logic above.
             using (response)
             {
-                FeatureFlagCircuitBreaker.RecordResponseReceived();
+                if (isHalfOpenProbe || response.IsSuccessStatusCode)
+                {
+                    circuitBreaker.RecordResponseReceived();
+                }
 
                 await response.EnsureSuccessfulApiCall(cancellationToken);
 
@@ -132,9 +148,6 @@ internal static class HttpClientExtensions
             }
         }
     }
-
-    internal static void ResetFeatureFlagCircuitBreakerForTests()
-        => FeatureFlagCircuitBreaker.Reset();
 
     static bool IsRetryableFlagsHttpRequestException(HttpRequestException exception)
     {
@@ -469,26 +482,6 @@ sealed class FeatureFlagRequestCircuitBreaker
     }
 
     public void RecordResponseReceived()
-    {
-        lock (_lock)
-        {
-            _state = State.Closed;
-            _consecutiveFailures = 0;
-            _openUntil = default;
-        }
-    }
-
-    public void RecordNonTransientFailure(bool isHalfOpenProbe)
-    {
-        if (!isHalfOpenProbe)
-        {
-            return;
-        }
-
-        RecordResponseReceived();
-    }
-
-    public void Reset()
     {
         lock (_lock)
         {
--- a/src/PostHog/Library/HttpClientExtensions.cs
+++ b/src/PostHog/Library/HttpClientExtensions.cs
@@ -13,6 +13,9 @@ namespace PostHog.Library;
 /// </summary>
 internal static class HttpClientExtensions
 {
+    static readonly TimeSpan FeatureFlagInitialRetryDelay = TimeSpan.FromMilliseconds(300);
+    static readonly FeatureFlagRequestCircuitBreaker FeatureFlagCircuitBreaker = new();
+
     /// <summary>
     /// Sends a POST request to the specified Uri containing the value serialized as JSON in the request body.
     /// Returns the response body deserialized as <typeparamref name="TBody"/>.
@@ -56,9 +59,14 @@ internal static class HttpClientExtensions
         CancellationToken cancellationToken)
     {
         var maxRetries = options.FeatureFlagRequestMaxRetries;
-        var currentDelay = options.InitialRetryDelay;
+        var currentDelay = FeatureFlagInitialRetryDelay;
         var maxDelay = options.MaxRetryDelay;
         var attempt = 0;
+
+        if (!FeatureFlagCircuitBreaker.TryEnter(timeProvider, out var isHalfOpenProbe))
+        {
+            throw new HttpRequestException("Feature flag request circuit breaker is open.");
+        }
 
         while (true)
         {
@@ -73,17 +81,39 @@ internal static class HttpClientExtensions
                     JsonSerializerHelper.Options,
                     cancellationToken);
             }
-            catch (HttpRequestException e) when (attempt <= maxRetries && IsRetryableFlagsHttpRequestException(e))
+            catch (HttpRequestException e) when (IsRetryableFlagsHttpRequestException(e))
             {
+                if (attempt > maxRetries ||
+                    !FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe))
+                {
+                    throw;
+                }
+
                 await Delay(timeProvider, currentDelay > maxDelay ? maxDelay : currentDelay, cancellationToken);
                 currentDelay = DoubleWithCap(currentDelay, maxDelay);
                 continue;
             }
-            catch (TaskCanceledException) when (!cancellationToken.IsCancellationRequested && attempt <= maxRetries)
+            catch (HttpRequestException)
             {
+                FeatureFlagCircuitBreaker.RecordNonTransientFailure(isHalfOpenProbe);
+                throw;
+            }
+            catch (TaskCanceledException) when (!cancellationToken.IsCancellationRequested)
+            {
+                if (attempt > maxRetries ||
+                    !FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe))
+                {
+                    throw;
+                }
+
                 await Delay(timeProvider, currentDelay > maxDelay ? maxDelay : currentDelay, cancellationToken);
                 currentDelay = DoubleWithCap(currentDelay, maxDelay);
                 continue;
+            }
+            catch (Exception) when (isHalfOpenProbe)
+            {
+                FeatureFlagCircuitBreaker.RecordNonTransientFailure(isHalfOpenProbe);
+                throw;
             }
 
             // Response processing is outside the try-catch so that exceptions from
@@ -91,6 +121,8 @@ internal static class HttpClientExtensions
             // be caught by the retry logic above.
             using (response)
             {
+                FeatureFlagCircuitBreaker.RecordResponseReceived();
+
                 await response.EnsureSuccessfulApiCall(cancellationToken);
 
                 var result = await response.Content.ReadAsStreamAsync(cancellationToken);
@@ -100,6 +132,9 @@ internal static class HttpClientExtensions
             }
         }
     }
+
+    internal static void ResetFeatureFlagCircuitBreakerForTests()
+        => FeatureFlagCircuitBreaker.Reset();
 
     static bool IsRetryableFlagsHttpRequestException(HttpRequestException exception)
     {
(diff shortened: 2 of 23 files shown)


marandaneto:
@haacked wdyt about this circuit break approach? (we can also make the 30s configurable)

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -13,6 +13,9 @@ namespace PostHog.Library;
 /// </summary>
 internal static class HttpClientExtensions
 {
+    static readonly TimeSpan FeatureFlagInitialRetryDelay = TimeSpan.FromMilliseconds(300);
+    static readonly FeatureFlagRequestCircuitBreaker FeatureFlagCircuitBreaker = new();
```
haacked: `suggestion`: This breaker is a `static` field, so one instance is shared by every `PostHogClient` in the process, and it keys on nothing (not host, token, or `HttpClient`). A connectivity blip on one client opens the circuit for all of them: a US-region outage makes EU `/flags` calls fail fast with "circuit breaker is open", and one project token's transient errors block every other token in the same process. Multiple clients in one process is a supported setup (US + EU, multiple tokens, or `new PostHogClient(...)`).

Scoping it per client keeps the cross-thread coordination you want and drops the cross-region coupling: make the breaker an instance field on `PostHogApiClient` (already built per `IOptions<PostHogOptions>`, so per host and token) and pass it into `PostJsonWithNetworkRetryAsync`. The only production caller is [`PostHogApiClient.cs:145`](https://github.com/PostHog/posthog-dotnet/blob/f073e4c9a5ea9e9122e9f4cd570fe809a6501a2f/src/PostHog/Api/PostHogApiClient.cs#L145). It also lets you delete the test-only `ResetFeatureFlagCircuitBreakerForTests` hook, since each test would get a fresh breaker.

tests/UnitTests/Library/HttpClientExtensionsTests.cs
```diff
@@ -554,14 +554,18 @@ public class ThePostJsonWithNetworkRetryAsyncMethod
...
-    static PostHogOptions CreateOptions(int maxRetries = 3) => new()
+    static PostHogOptions CreateOptions(int maxRetries = 3, TimeSpan? maxRetryDelay = null)
     {
-        ProjectToken = "test-api-key",
-        MaxRetries = maxRetries,
-        FeatureFlagRequestMaxRetries = maxRetries,
-        InitialRetryDelay = TimeSpan.FromMilliseconds(1),
-        MaxRetryDelay = TimeSpan.FromSeconds(30)
-    };
+        HttpClientExtensions.ResetFeatureFlagCircuitBreakerForTests();
```
haacked: `blocking`: These tests can flake in CI. The breaker is a process-global `static`, and resetting it here only isolates this test class from itself. xUnit runs test classes in parallel (no `xunit.runner.json` or `[Collection]` in this assembly), and `FeatureFlagsTests`, `FeatureFlagEvaluationsTests`, `FeatureFlagExtensionsTests`, `ETagSupportTests`, and `PostHogClientTests` all issue real `/flags` requests through the same shared breaker without resetting it.

While `OpensCircuitAfterConsecutiveTransientFailuresAndFailsFast` accumulates failures across its `await` points, a successful `/flags` request from any of those classes calls `RecordResponseReceived` and zeroes `_consecutiveFailures`, so the count never reaches 5 and `Assert.Equal(5, handler.RequestCount)` fails. The same breaker also means a concurrent request from another test can hit the open circuit and throw "circuit breaker is open" when it expected a flags response.

The fix is to make the breaker instance-scoped (per client) so each test gets its own. Serializing with a `[Collection]` isn't enough on its own, since it would have to cover every `/flags`-touching class.

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -73,24 +81,48 @@ internal static class HttpClientExtensions
                     JsonSerializerHelper.Options,
                     cancellationToken);
             }
-            catch (HttpRequestException e) when (attempt <= maxRetries && IsRetryableFlagsHttpRequestException(e))
+            catch (HttpRequestException e) when (IsRetryableFlagsHttpRequestException(e))
             {
+                if (attempt > maxRetries ||
+                    !FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe))
```
haacked: `suggestion`: With `FeatureFlagRequestMaxRetries = 0`, the circuit never opens, even during a total `/flags` outage. Because `||` short-circuits, when `attempt > maxRetries` is true `RecordTransientFailure` is never called, so the failure that fails the whole call goes uncounted. At `maxRetries = 0` attempt 1 throws before recording anything; at the default `maxRetries = 1` each failed call records just one failure.

Record the failure before the retry decision to close the `maxRetries = 0` hole:

```csharp
catch (HttpRequestException e) when (IsRetryableFlagsHttpRequestException(e))
{
    var circuitClosed = FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe);
    if (attempt > maxRetries || !circuitClosed)
    {
        throw;
    }
    ...
}
```

This change also applies to the `TaskCanceledException` timeout block. Note that this makes one call count every attempt, so the breaker trips after fewer calls. Decide whether the threshold should count attempts or whole calls.

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -56,10 +59,15 @@ internal static class HttpClientExtensions
         CancellationToken cancellationToken)
     {
         var maxRetries = options.FeatureFlagRequestMaxRetries;
-        var currentDelay = options.InitialRetryDelay;
+        var currentDelay = FeatureFlagInitialRetryDelay;
```
haacked: `question`: This hardcodes the flags first-retry delay to 300ms and stops reading `options.InitialRetryDelay`. The batch path (`PostJsonWithRetryAsync`) still honors `options.InitialRetryDelay`, so the same option now drives one retry path and is silently ignored by the other. A user who raises `InitialRetryDelay` to slow flag retries would still get 300ms.

Is dropping the option for the flags path intentional, or should it stay `options.InitialRetryDelay` (or get its own option)?

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -73,24 +81,48 @@ internal static class HttpClientExtensions
...
+                    !FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe))
+                {
+                    throw;
+                }
+
                 await Delay(timeProvider, currentDelay > maxDelay ? maxDelay : currentDelay, cancellationToken);
                 currentDelay = DoubleWithCap(currentDelay, maxDelay);
                 continue;
             }
+            catch (Exception) when (isHalfOpenProbe)
```
haacked, line 133: `suggestion`: A user-cancelled probe is treated as recovery and drops the cooldown. When the cooldown just expired and this thread is the single half-open probe, if the caller's token fires during the request the `TaskCanceledException` has `IsCancellationRequested == true`, so the filter above doesn't match and this `catch (Exception) when (isHalfOpenProbe)` runs `RecordNonTransientFailure(true)` then `RecordResponseReceived()`, closing the circuit. A cancellation says nothing about service health, but the next calls now flow through and have to re-accumulate failures during a real outage.

Narrow window, small fix. Catch the cancelled probe first and re-arm the cooldown:

```csharp
catch (OperationCanceledException) when (isHalfOpenProbe && cancellationToken.IsCancellationRequested)
{
    FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe: true);
    throw;
}
```

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -73,24 +81,48 @@ internal static class HttpClientExtensions
...
+                FeatureFlagCircuitBreaker.RecordNonTransientFailure(isHalfOpenProbe);
+                throw;
+            }
 
             // Response processing is outside the try-catch so that exceptions from
             // EnsureSuccessfulApiCall (which may return HttpRequestException for 404s) won't
             // be caught by the retry logic above.
             using (response)
             {
+                FeatureFlagCircuitBreaker.RecordResponseReceived();
```
haacked: `question`: `RecordResponseReceived()` runs on any response before `EnsureSuccessfulApiCall`, so a 5xx zeroes `_consecutiveFailures`. Resetting on a status failure matches "status failures don't trip the circuit", but it also means a `/flags` endpoint flapping between connection resets and 5xx never opens the breaker, since each 5xx clears the transient count mid-accumulation.

Is resetting on any non-2xx intentional, or should the counter only reset on 2xx?

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -385,4 +420,95 @@ public static async Task EnsureSuccessfulApiCall(
...
+    {
+        lock (_lock)
+        {
+            _state = State.Closed;
+            _consecutiveFailures = 0;
+            _openUntil = default;
+        }
+    }
+
+    public void RecordNonTransientFailure(bool isHalfOpenProbe)
```
haacked: `suggestion`: `RecordNonTransientFailure` reads like it records a failure, but it closes the circuit (resets to `Closed`) and only when called during a probe; otherwise it's a no-op. Anyone reading this code later will misread what it does. The real intent is that a non-transient response means the endpoint is reachable, so a probe should close the circuit.

Inline that at the two call sites and drop the method:

```csharp
// in the non-transient HttpRequestException catch
if (isHalfOpenProbe)
{
    FeatureFlagCircuitBreaker.RecordResponseReceived();
}
throw;
```

tests/UnitTests/Library/HttpClientExtensionsTests.cs
```diff
@@ -644,13 +650,93 @@ public async Task ThrowsAfterMaxRetriesWhenConnectionResetPersists()
...
+                new { api_key = "test", distinct_id = "user-1" },
+                timeProvider,
+                options,
+                CancellationToken.None));
+
+        Assert.Equal(5, handler.RequestCount);
+    }
+
+    [Fact]
+    public async Task HalfOpenProbeClosesCircuitAfterCooldown()
```
haacked, line 843: `suggestion`: No test covers the half-open probe failing and re-opening the circuit. `HalfOpenProbeClosesCircuitAfterCooldown` only exercises the probe succeeding, but the `isHalfOpenProbe` branch in `RecordTransientFailure` is what re-opens the breaker when the probe fails; if that regressed, a failed probe would let traffic back through to a down server.

Mirror this test but queue a 6th connection-reset exception instead of the OK response, then assert the probe call throws and the next call fails fast with `RequestCount` unchanged.

haacked (approved):

haacked:
@marandaneto I meant to submit the review as a comment, not approval! But I trust you to do the right thing. :smile:

marandaneto:
> @marandaneto I meant to submit the review as a comment, not approval! But I trust you to do the right thing. 😄

thx for the review, yeah it was just a draft with a concept, will check and work on it on monday, then i can port this change across other sdks that have a similar situation

marandaneto:
i re-requested review to remove the approval, but no need to re-review until i address the comments and make it ready for review

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -73,24 +81,48 @@ internal static class HttpClientExtensions
...
             {
+                if (attempt > maxRetries ||
+                    !FeatureFlagCircuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe))
+                {
+                    throw;
+                }
+
                 await Delay(timeProvider, currentDelay > maxDelay ? maxDelay : currentDelay, cancellationToken);
                 currentDelay = DoubleWithCap(currentDelay, maxDelay);
                 continue;
```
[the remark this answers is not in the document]
marandaneto: Fixed in da75990: transient failures are now recorded before the retry-budget check for both HttpRequestException and timeout paths, so maxRetries=0 and half-open probe failures cannot leave the breaker stuck half-open. Added maxRetries=0 coverage.

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -73,24 +81,48 @@ internal static class HttpClientExtensions
...
             }
+            catch (Exception) when (isHalfOpenProbe)
+            {
+                FeatureFlagCircuitBreaker.RecordNonTransientFailure(isHalfOpenProbe);
+                throw;
+            }
```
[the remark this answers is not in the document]
marandaneto: Fixed in da75990: extracted the duplicated retry-delay/update logic into a local DelayBeforeRetry helper.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in da75990: the breaker is now instance-scoped on PostHogApiClient and passed into PostJsonWithNetworkRetryAsync, so it coordinates concurrent calls per client without coupling different hosts/tokens.

tests/UnitTests/Library/HttpClientExtensionsTests.cs
marandaneto: Fixed in da75990: removed the process-global static breaker/test reset hook. Production uses a per-PostHogApiClient breaker, and tests create isolated breaker instances where needed.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in da75990: RecordTransientFailure now runs before the retry decision in both transient exception paths. I chose to count failed attempts, including terminal attempts when maxRetries=0, so a total outage still opens the circuit.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in da75990: user-cancelled half-open probes now re-arm the cooldown instead of closing the circuit as recovered.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in da75990: non-2xx responses no longer reset the transient-failure counter unless they are the half-open probe response. Normal closed-state reset only happens on 2xx success, so interleaved 5xx responses won't mask transport outages.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in da75990: removed RecordNonTransientFailure and inlined the half-open probe recovery behavior at the call sites so the intent is explicit.

tests/UnitTests/Library/HttpClientExtensionsTests.cs
marandaneto: Fixed in da75990: added HalfOpenProbeReopensCircuitWhenItFails, which queues a failed probe and verifies the following call fails fast without another request.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Changed in 346bcd1: feature-flag retries now use options.InitialRetryDelay as the first delay and continue doubling via DoubleWithCap for later retries. Added a focused test that sets InitialRetryDelay=10ms and verifies the first retry waits 10ms and the second waits 20ms.

marandaneto:
@PostHog/team-client-libraries still pending here

.changeset/quiet-flags-retry.md
```diff
@@ -0,0 +1,5 @@
+---
+'PostHog': patch
+---
+
+Retry feature flag requests after transient network errors only. The feature flag request retry count defaults to 1 and can be set to 0 to disable retries.
```
ioannisj: Stale changeset from https://github.com/PostHog/posthog-dotnet/pull/245?

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -55,12 +55,25 @@ internal static class HttpClientExtensions
...
         PostHogOptions options,
-        CancellationToken cancellationToken)
+        CancellationToken cancellationToken,
+        FeatureFlagRequestCircuitBreaker? circuitBreaker = null)
     {
         var maxRetries = options.FeatureFlagRequestMaxRetries;
         var currentDelay = options.InitialRetryDelay;
         var maxDelay = options.MaxRetryDelay;
         var attempt = 0;
+        circuitBreaker ??= new FeatureFlagRequestCircuitBreaker();
```
ioannisj: This creates a new instance per call when param is omitted. Should it be a required param?

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -418,4 +464,75 @@ public static async Task EnsureSuccessfulApiCall(
...
         throw await CreateApiException(response, cancellationToken);
     }
-}
\ No newline at end of file
+}
+
+sealed class FeatureFlagRequestCircuitBreaker
+{
+    const int FailureThreshold = 5;
+    static readonly TimeSpan OpenDuration = TimeSpan.FromSeconds(30);
```
ioannisj, line 479: Should these be configurable you think?

ioannisj (commented):
LG, left a couple of comments

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: its hardcoded 30s in other sdks as well, lets keep as is until we see its needed

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Agreed — made the circuit breaker a required parameter in af30ef2, so callers cannot accidentally create a new breaker per call.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: We decided not to make these configurable for now; keeping the threshold/duration internal avoids adding more public options unless we see a need to tune them.

ioannisj (approved):

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -75,24 +87,57 @@ internal static class HttpClientExtensions
...
+                if (attempt > maxRetries || !circuitClosed)
+                {
+                    throw;
+                }
+
+                await DelayBeforeRetry();
                 continue;
             }
-            catch (TaskCanceledException) when (!cancellationToken.IsCancellationRequested && attempt <= maxRetries)
+            catch (HttpRequestException)
```
haacked, line 111: `suggestion`: When a half-open probe fails with a non-retryable transport error (connection refused, DNS failure, or a TLS handshake error), the circuit still closes. The non-retryable `catch (HttpRequestException)` at line 101 and the catch-all `catch (Exception) when (isHalfOpenProbe)` at [line 125](https://github.com/PostHog/posthog-dotnet/blob/1117640c2ed9180ccedc7b3a08702f66870a8e5b/src/PostHog/Library/HttpClientExtensions.cs#L125) both call `RecordResponseReceived()`, which sets the breaker back to `Closed`, treating a still-unreachable server as if it had recovered even though no response ever came back.

If the endpoint drops connections during the outage but starts refusing them by the time the 30s cooldown ends, the probe hits this path and the circuit closes as if the service recovered. The next wave of concurrent callers floods back in and has to re-accumulate 5 failures from scratch.

Reopen the circuit instead of closing on these paths:

```suggestion
            catch (HttpRequestException)
            {
                if (isHalfOpenProbe)
                {
                    circuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe: true);
                }
                throw;
            }
```

(The `catch (Exception) when (isHalfOpenProbe)` block a few lines down needs the same change: call `RecordTransientFailure(timeProvider, isHalfOpenProbe: true)` instead of `RecordResponseReceived()`.)

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -418,4 +463,75 @@ public static async Task EnsureSuccessfulApiCall(
...
+            {
+                Open(timeProvider);
+                return false;
+            }
+
+            return true;
+        }
+    }
+
+    public void RecordResponseReceived()
```
haacked: `suggestion`: `RecordResponseReceived()` reads as if a response came back, but two of its three call sites fire from exception catches where nothing was received, treating a failure as if it were recovery. Rename it to `Close()` so the success path, the non-transient error catch, and the probe catch-all all read as the state transition they perform rather than an event that didn't happen.

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -75,24 +87,57 @@ internal static class HttpClientExtensions
                     JsonSerializerHelper.Options,
                     cancellationToken);
             }
-            catch (HttpRequestException e) when (attempt <= maxRetries && IsRetryableFlagsHttpRequestException(e))
+            catch (HttpRequestException e) when (IsRetryableFlagsHttpRequestException(e))
```
haacked, line 102: `suggestion`: The retryable catches for `HttpRequestException` and `TaskCanceledException` now run identical bodies: record the transient failure, throw if retries are exhausted or the circuit tripped, otherwise delay and continue. `DelayBeforeRetry` extracted the delay into a local function, but the record-and-decide logic is still duplicated.

Fold it into one local function:

```csharp
async Task<bool> ShouldRetryAfterTransientFailure()
{
    var circuitClosed = circuitBreaker.RecordTransientFailure(timeProvider, isHalfOpenProbe);
    if (attempt > maxRetries || !circuitClosed)
    {
        return false;
    }
    await DelayBeforeRetry();
    return true;
}
```

Each catch becomes: `if (!await ShouldRetryAfterTransientFailure()) { throw; } continue;`

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -55,13 +55,25 @@ internal static class HttpClientExtensions
...
         PostHogOptions options,
+        FeatureFlagRequestCircuitBreaker circuitBreaker,
         CancellationToken cancellationToken)
     {
         var maxRetries = options.FeatureFlagRequestMaxRetries;
         var currentDelay = options.InitialRetryDelay;
         var maxDelay = options.MaxRetryDelay;
         var attempt = 0;
 
+        if (!circuitBreaker.TryEnter(timeProvider, out var isHalfOpenProbe))
```
haacked, line 66: `suggestion`: The circuit breaker's state transitions are decided by the caller instead of the breaker itself. Adding the breaker turned one exception path into five catch clauses, three of which exist only to record a probe or response outcome, and `RecordResponseReceived()` is now called from three separate places. The rule "we reached the server, so close the circuit" lives in the caller's catch filters, which is exactly what let the probe-closing bug flagged elsewhere in this review slip through.

Consider having the retry loop report a single outcome per attempt (success, transient failure, or reached-the-server) and letting `FeatureFlagRequestCircuitBreaker` own the Closed/Open/HalfOpen transitions from that one signal. Fine to defer while this is a prototype, but worth doing before this ships since the file already churns on this exact loop.

tests/UnitTests/Library/HttpClientExtensionsTests.cs
```diff
@@ -639,18 +679,193 @@ public async Task ThrowsAfterMaxRetriesWhenConnectionResetPersists()
...
+                new { api_key = "test", distinct_id = "user-1" },
+                timeProvider,
+                options,
+                circuitBreaker,
+                CancellationToken.None));
+        Assert.Equal(5, handler.RequestCount);
+    }
+
+    [Fact]
+    public async Task HalfOpenProbeClosesCircuitAfterCooldown()
```
haacked, line 843: `suggestion`: Nothing tests that only one probe is allowed through during recovery. `TryEnter`'s `return _state != State.HalfOpen;` is what rejects a second caller while a probe is in flight, and no test exercises it, so a regression that let a thundering herd of probes hit a recovering endpoint would ship green.

It's directly testable on the class without threads:

```csharp
var breaker = new FeatureFlagRequestCircuitBreaker();
var timeProvider = new FakeTimeProvider();
for (var i = 0; i < 5; i++) breaker.RecordTransientFailure(timeProvider, isHalfOpenProbe: false);
timeProvider.Advance(TimeSpan.FromSeconds(31));
Assert.True(breaker.TryEnter(timeProvider, out var firstProbe));
Assert.True(firstProbe);
Assert.False(breaker.TryEnter(timeProvider, out var secondProbe));
Assert.False(secondProbe);
```

tests/UnitTests/Library/HttpClientExtensionsTests.cs
```diff
@@ -581,16 +584,52 @@ public async Task RetriesOnConnectionResetThenSucceeds()
             new { api_key = "test", distinct_id = "user-1" },
             timeProvider,
             options,
+            new FeatureFlagRequestCircuitBreaker(),
             CancellationToken.None);
 
         await handler.WaitForRequestCountAsync(1);
-        timeProvider.Advance(TimeSpan.FromSeconds(1));
+        Assert.Equal(1, handler.RequestCount);
```
haacked, line 591: `suggestion`: No test confirms a success resets `_consecutiveFailures`, so "consecutive" in the failure count is unverified. `RetriesOnConnectionResetThenSucceeds` only has one failure before the success, nowhere near the threshold of 5, so a regression that counted cumulative failures instead of consecutive ones would still pass while opening the circuit under ordinary intermittent errors in production.

Add a test that runs 4 failures, a success, then 4 more failures across separate calls and asserts the circuit stays closed the whole time. Mirror the structure of `OpensCircuitAfterConsecutiveTransientFailuresWhenRetriesAreDisabled`.

haacked (commented):
Nice work resolving the earlier feedback! A few more non-blocking suggestions inline.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Good catch — fixed in 4ad8cc8. Half-open probes now re-open the circuit on non-retryable HttpRequestException and unexpected exceptions instead of closing it as recovered.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in 4ad8cc8: renamed RecordResponseReceived to Close, and removed the exception-path calls that were closing without a response.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in 4ad8cc8: folded the duplicate transient failure retry decision into ShouldRetryAfterTransientFailure().

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Addressed the risky part in 4ad8cc8 by centralizing the transient-failure retry decision and making half-open failure paths consistently re-open the circuit. I kept the broader state-machine refactor out of this PR to avoid expanding scope.

tests/UnitTests/Library/HttpClientExtensionsTests.cs
marandaneto: Fixed in 4ad8cc8: added AllowsOnlyOneHalfOpenProbeDuringRecovery to verify a second caller is rejected while a half-open probe is in flight.

tests/UnitTests/Library/HttpClientExtensionsTests.cs
marandaneto: Fixed in 4ad8cc8: added SuccessfulResponseResetsConsecutiveFailures to verify a success resets the consecutive transient failure counter.

marandaneto:
will wait for people to upgrade/test/complain (hopefully not) this approach before i add to the sdk test harness/spec/port to other SDKs

PR #249, opened 2026-06-27
fix: fall back when payload compression fails

marandaneto:

If local gzip compression fails while preparing a batch upload, the SDK should still send the original payload uncompressed instead of failing before the request is sent.

This updates the compressed batch path to build gzip content first and fall back to the normal JSON request path when gzip content creation fails.

- `dotnet test tests/UnitTests/UnitTests.csproj --filter HttpClientExtensionsTests`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented as part of a cross-SDK consistency pass for client-side compression failure fallback. Kept the change surgical and limited to the gzip request-construction path.

### DIFF
--- a/.changeset/cold-buckets-gzip.md
+++ b/.changeset/cold-buckets-gzip.md
@@ -1,0 +1,5 @@
+---
+'posthog-dotnet': patch
+---
+
+Fall back to uncompressed batch uploads when local gzip compression fails.
--- a/.changeset/cold-buckets-gzip.md
+++ b/.changeset/cold-buckets-gzip.md
@@ -1,5 +1,6 @@
 ---
-'posthog-dotnet': patch
+'PostHog': patch
+'PostHog.AspNetCore': patch
 ---
 
 Fall back to uncompressed batch uploads when local gzip compression fails.
--- a/src/PostHog/Library/HttpClientExtensions.cs
+++ b/src/PostHog/Library/HttpClientExtensions.cs
@@ -13,7 +13,7 @@ namespace PostHog.Library;
 /// </summary>
 internal static class HttpClientExtensions
 {
-    internal static Func<byte[], ByteArrayContent> CreateCompressedJsonContent = CreateGzipJsonContent;
+    internal static Func<object, CancellationToken, Task<ByteArrayContent>> CreateCompressedJsonContentAsync = CreateGzipJsonContentAsync;
     /// <summary>
     /// Sends a POST request to the specified Uri containing the value serialized as JSON in the request body.
     /// Returns the response body deserialized as <typeparamref name="TBody"/>.
@@ -276,8 +276,7 @@ internal static class HttpClientExtensions
         object content,
         CancellationToken cancellationToken)
     {
-        var jsonBytes = JsonSerializer.SerializeToUtf8Bytes(content, JsonSerializerHelper.Options);
-        var compressedContent = TryCreateCompressedJsonContent(jsonBytes);
+        var compressedContent = await TryCreateCompressedJsonContentAsync(content, cancellationToken);
         if (compressedContent is null)
         {
             return await httpClient.PostAsJsonAsync(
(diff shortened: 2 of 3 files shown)


src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -268,30 +268,57 @@ static Task Delay(TimeProvider timeProvider, TimeSpan delay, CancellationToken c
...
+            return compressedContent;
+        }
+        catch (IOException)
+        {
+            return null;
+        }
+        catch (InvalidDataException)
+        {
+            return null;
         }
```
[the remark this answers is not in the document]
marandaneto: Fixed in 8062488 — added a regression test that forces gzip compression to fail and verifies the request is sent uncompressed.

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -268,30 +268,57 @@ static Task Delay(TimeProvider timeProvider, TimeSpan delay, CancellationToken c
...
     }
```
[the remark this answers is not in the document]
marandaneto: Fixed in 8062488 — added debug-level observability when gzip compression fails before falling back.

src/PostHog/Library/HttpClientExtensions.cs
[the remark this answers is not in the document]
marandaneto: Fixed in 8062488 — broadened the local compression-failure handling to realistic write failure types and covered the path with a test.

.changeset/cold-buckets-gzip.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-dotnet': patch
```
marandaneto: it should be `PostHog`

marandaneto: also add the aspnet changeset

marandaneto: Fixed in 845f0a0 — renamed the changeset package to `PostHog` and added a patch changeset entry for `PostHog.AspNetCore`.

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -268,31 +270,58 @@ static Task Delay(TimeProvider timeProvider, TimeSpan delay, CancellationToken c
...
+        }
+    }
+
+    static ByteArrayContent? TryCreateCompressedJsonContent(byte[] jsonBytes)
+    {
+        try
+        {
+            return CreateCompressedJsonContent(jsonBytes);
+        }
+        catch (Exception ex) when (ex is IOException or InvalidDataException or NotSupportedException or ObjectDisposedException)
```
ioannisj, line 303: Should we also test for the other 3 exception types here?

src/PostHog/Library/HttpClientExtensions.cs
```diff
@@ -268,31 +270,58 @@ static Task Delay(TimeProvider timeProvider, TimeSpan delay, CancellationToken c
...
+        }
+    }
+
+    static ByteArrayContent CreateGzipJsonContent(byte[] jsonBytes)
+    {
         using var memoryStream = new MemoryStream(4096);
         using (var gzipStream = new GZipStream(memoryStream, CompressionLevel.Fastest, leaveOpen: true))
         {
-            await JsonSerializer.SerializeAsync(gzipStream, content, JsonSerializerHelper.Options, cancellationToken);
+            gzipStream.Write(jsonBytes, 0, jsonBytes.Length);
```
ioannisj: shouldn't we pass cancellationToken along? Seems like a regression from original

ioannisj (approved):
LG but left a couple of comments

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in f5626c9 — converted the fallback regression test to cover all handled compression exceptions: IOException, InvalidDataException, NotSupportedException, and ObjectDisposedException.

src/PostHog/Library/HttpClientExtensions.cs
marandaneto: Fixed in f5626c9 — restored the async gzip serialization path so the cancellationToken is passed through to JsonSerializer.SerializeAsync again.

marandaneto: good call, thjx

PR #250, opened 2026-06-29
fix(feature-management): enumerate flags without PersonalApiKey

### LINKED ISSUE #64: bug: Feature Management doesn't work with local evaluation
If you don't provide your PersonalApiKey, then feature management is broken. We should allow falling back to `/decide` in that case.

jojosenthusiast:
Fixes #64

## Summary
- Keeps local evaluation as the source when a PersonalApiKey is available.
- Falls back to `/flags` with a stable sentinel distinct id for feature enumeration without a PersonalApiKey.
- Adds unit coverage for both `PostHogFeatureDefinitionProvider` and `PostHogVariantFeatureManager`.

## Verification
- [x] `dotnet test tests/UnitTests.AspNetCore --filter FullyQualifiedName~Fallback -m:1`
- [x] `dotnet test tests/UnitTests.AspNetCore -m:1`
- [x] `git diff --check main...HEAD`

## Risk
Medium. This affects feature flag enumeration only; no public API changes are added.

### DIFF
--- a/src/PostHog.AspNetCore/FeatureManagement/FeatureEnumerationFallback.cs
+++ b/src/PostHog.AspNetCore/FeatureManagement/FeatureEnumerationFallback.cs
@@ -1,0 +1,38 @@
+namespace PostHog.FeatureManagement;
+
+/// <summary>
+/// Shared fallback used by <see cref="PostHogFeatureDefinitionProvider"/> and
+/// <see cref="PostHogVariantFeatureManager"/> to enumerate flag keys when no
+/// <see cref="Config.PostHogOptions.PersonalApiKey"/> is configured (no local evaluation source).
+/// </summary>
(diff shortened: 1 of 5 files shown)


turnipdabeets:
@jojosenthusiast thanks! LGTM, I think the reptile comment is worth addressing before merging https://github.com/PostHog/posthog-dotnet/pull/250#discussion_r3492388656. Mind giving that a look? 

I think you'll also need to sign your commits to merge the PR.

jojosenthusiast:
I looked at the Greptile bit, removed the duplicated test constants, and both tests now use the production sentinel directly. Re-ran the fallback tests and the AspNetCore unit tests locally, so this should be good to go now.

turnipdabeets (approved):

PR #253, opened 2026-07-01
ci: skip heavy CI for non-impacting changes

marandaneto:
## Summary
- replace the required CI detector with a local conservative generic safe-file classifier for PR `filename` and `previous_filename` entries
- keep required check names running, but bail out of heavy .NET build/package/test/format steps when every changed path is in the global safe set
- treat pushes, API failures, empty changed-file lists, source/test/build/package/config/lockfile changes, current CI workflows, and unknown files as full-CI changes
- add path filters to optional SDK compliance and CodeQL workflows

## Validation
- PyYAML and Ruby YAML parse for `.github/workflows/main.yaml`
- `bash -n` on the embedded classifier script
- classifier sample checks for markdown, docs/templates/CODEOWNERS/dependabot, allowed auxiliary workflows, source files, lockfiles, current CI workflow, unsafe previous filenames, push events, API failures, and empty changed-file lists
- `jq` sample confirmed the PR files API expression emits both `filename` and non-empty `previous_filename` values


marandaneto:
Closing this for now — I'll investigate other options. We don't want hardcoded paths this way.

PR #257, opened 2026-07-02
feat: add SecretKey config, deprecate PersonalApiKey

turnipdabeets:
## Problem

The credential used for local feature flag evaluation and remote config accepts **either** a Personal API Key (`phx_...`) **or** a Project Secret API Key (`phs_...`). The option that carries it is named `PersonalApiKey`, which is confusing — it implies only a personal key is valid.

## Change

Add a new canonical `PostHogOptions.SecretKey` property and keep `PersonalApiKey` as a **deprecated alias**. Non-breaking.

- `SecretKey` accepts a Personal API Key or a Project Secret API Key.
- `PersonalApiKey` is marked `[Obsolete("Use SecretKey instead...")]` (warning, not error).
- Both are backed by separate fields; the effective value resolves to `SecretKey` when non-empty, otherwise `PersonalApiKey`. When both are set, `SecretKey` wins. This mirrors the existing `ProjectToken`/`ProjectApiKey` alias pattern in this file, so all internal consumers read the canonical property and never touch the obsolete member (no obsolete-warning noise under `TreatWarningsAsErrors`).
- XML docs, user-secrets example, exception messages, and `<see cref>` references updated to `SecretKey`.
- Added a changeset (`PostHog`, minor).

## Context

Coordinated rename across the backend SDKs (Slack discussion). Tracks the rename item in PostHog/posthog-js#4046. Mirrors the draft in posthog-python (#727). This PR covers only the rename item.

### DIFF
--- a/.changeset/secret-key-config.md
+++ b/.changeset/secret-key-config.md
@@ -1,0 +1,5 @@
+---
+'PostHog': minor
+---
+
+Add `PostHogOptions.SecretKey` for local feature flag evaluation and remote config. It accepts either a Personal API Key (`phx_...`) or a Project Secret API Key (`phs_...`). The existing `PersonalApiKey` option is now a deprecated alias; when both are set, `SecretKey` takes precedence.
--- a/samples/HogTied.Web/Pages/Index.cshtml
+++ b/samples/HogTied.Web/Pages/Index.cshtml
@@ -47,17 +47,17 @@
                 </strong>
             </div>
 
-            if (Model.PersonalApiKeyIsSet) {
+            if (Model.SecretKeyIsSet) {
                 <div class="alert alert-info mt-3" role="alert">
                     <strong>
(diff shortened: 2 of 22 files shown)


turnipdabeets:
**Marked ready for review.**

Naming is settled — this is a generic **`secret_key`** config that accepts **either** a Personal API Key or a Project Secret API Key (both coexist; same `Authorization: Bearer` header). `personal_api_key` stays as a deprecated **config-name** alias that forwards into it — *not* a deprecation of personal keys as a credential. It's a name-level change and is crash-safe with respect to using the key (adds no new error path).

Merge timing still coordinates with the project-secret-key rollout (feature-flagged today) — ready-for-review ≠ merge-now.

dustinbyrne (approved):
examples still contain references to `POSTHOG_PERSONAL_API_KEY`/`personal_api_key`

PR #260, opened 2026-07-05
feat: add before send callback

marandaneto:

Customers need a configuration hook to inspect, modify, or drop events before upload, matching the before-send behavior available in other PostHog SDKs.

This adds `PostHogOptions.BeforeSend`, invoked after event enrichment (SDK properties, super properties, `$is_server`, groups, and feature flags) and before batch serialization. Returning `null` drops the event.

- `dotnet test tests/UnitTests/UnitTests.csproj --filter 'FullyQualifiedName~PostHogClientTests.TheCaptureMethod'`
- `dotnet build src/PostHog/PostHog.csproj`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the pi coding agent. The hook runs after the event is fully enriched by moving before-send processing into the batch resolution path, so feature flag enrichment is visible to the callback before serialization. A dropped event is filtered out of the outgoing batch.

### DIFF
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -68,7 +68,7 @@ public sealed class PostHogClient : IPostHogClient
             loggerFactory.CreateLogger<PostHogApiClient>()
         );
         _asyncBatchHandler = new AsyncBatchHandler<CapturedEvent, CapturedEventBatchContext>(
-            batch => _apiClient.CaptureBatchAsync(batch, CancellationToken.None),
+            CaptureBatchAsync,
             batchContextFunc: () => new CapturedEventBatchContext(
                 new FallbackFeatureFlagCache(
                     new MemoryFeatureFlagCache(_timeProvider, 10000, 0.2),
@@ -331,7 +331,7 @@ public sealed class PostHogClient : IPostHogClient
         _logger.LogWarnCaptureFailed(eventName, capturedEvent.Properties.Count, _asyncBatchHandler.Count);
         return false;
 
-        async Task<CapturedEvent?> BatchTask(CapturedEventBatchContext context)
+        async Task<CapturedEvent> BatchTask(CapturedEventBatchContext context)
         {
             CapturedEvent enrichedEvent;
             if (flags is not null)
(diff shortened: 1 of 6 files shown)


tests/UnitTests/PostHogClientTests.cs
```diff
@@ -415,6 +415,60 @@ static JsonElement GetOnlyBatchItem(FakeHttpMessageHandler.RequestHandler batchH
...
+            options.BeforeSend = _ => null;
+        }));
+        var requestHandler = container.FakeHttpMessageHandler.AddBatchResponse();
+        var client = container.Activate<PostHogClient>();
+
+        Assert.True(client.Capture("test-user", "drop-me"));
+        await client.FlushAsync();
+
+        Assert.Empty(requestHandler.ReceivedRequests);
+    }
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in e01c090 — Added the throwing-callback case to the BeforeSend drop test.

src/PostHog/PostHogClient.cs
```diff
@@ -1278,6 +1334,18 @@ public async ValueTask DisposeAsync()
...
+        EventId = 22,
+        Level = LogLevel.Debug,
+        Message = "Event {EventName} was dropped by the before send callback")]
+    public static partial void LogDebugBeforeSendDropped(this ILogger<PostHogClient> logger, string eventName);
+
+    [LoggerMessage(
+        EventId = 23,
+        Level = LogLevel.Error,
+        Message = "Error in before send callback for event {EventName}; dropping event")]
+    public static partial void LogErrorBeforeSendException(this ILogger<PostHogClient> logger, Exception exception, string eventName);
```
ioannisj: There's already id 22 and 23 defined. These should be appended at the end of the scope since the messages here are declared sorted and updated to ids 25 and 26

ioannisj (approved):
LG, left a small comment

src/PostHog/PostHogClient.cs
marandaneto: Good catch — fixed in 9ab20d9 by moving these logger messages to the end of the scope and assigning EventIds 25 and 26.

PR #261, opened 2026-07-15
feat(flags): add $feature_flag_has_experiment to $feature_flag_called events

haacked:

Part of a cross-SDK effort to shrink `$feature_flag_called` events. This first phase adds a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, reflecting the server's `has_experiment` signal: `metadata.has_experiment` in the `/flags?v=2` response and `has_experiment` on `/api/feature_flag/local_evaluation` flag definitions. When the server does not report the field (older deployments, bootstrapped flags), the property is omitted, so it is tri-state: `true`, `false`, or absent (unknown).

This lets ingestion distinguish experiment-linked flag events (which need the full property set for exposure analysis) from the rest, and lets us measure the split before a later phase strips the expensive properties (`$feature/<key>`, `$feature_flag_payload`, per-event system metadata) from non-experiment flag events. This PR intentionally minimizes nothing: no properties are removed, no config options are added, and dedupe behavior is unchanged.

## Changes

- `FeatureFlagMetadata.HasExperiment` (`has_experiment`) on the `/flags` DTO and the same property on `LocalFeatureFlag` for `/local_evaluation`; both factories thread it onto the shared `FeatureFlag` record (included in record equality).
- Both send paths funnel through `BuildFeatureFlagCalledProperties`, which now always sets `$feature_flag_has_experiment`; synthetic missing/error flags report `false`. Legacy v3 responses and missing metadata safely default to `false`.
- `PublicAPI.Unshipped.txt` updated (RS0016); minor changeset.

`dotnet build PostHog.sln`: 0 errors, 0 warnings. `dotnet test`: 1,012 passed (UnitTests 944 + AspNetCore 49 + PostHog.AI 19), 0 failures. Two new theories cover true / false / absent from flags metadata and local evaluation, and 20 exact-JSON expectations were updated. `bin/fmt` clean.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/tall-cows-attend.md
+++ b/.changeset/tall-cows-attend.md
@@ -1,0 +1,5 @@
+---
+"PostHog": minor
+---
+
+Add a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, sourced from the server-reported `has_experiment` field on flag responses. Defaults to `false` when the server does not report the field.
--- a/.changeset/tall-cows-attend.md
+++ b/.changeset/tall-cows-attend.md
@@ -2,4 +2,4 @@
 "PostHog": minor
 ---
 
-Add a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, sourced from the server-reported `has_experiment` field on flag responses. Defaults to `false` when the server does not report the field.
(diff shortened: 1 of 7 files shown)


turnipdabeets (approved):
LGTM! Both send paths covered (flags metadata + local evaluation), null-safe defaults to false on older deployments, and the test matrix covers true/false/absent for both sources.

PR #267, opened 2026-07-30
fix: avoid mutating caller property dictionaries

### LINKED ISSUE #266: Capturing a event mutates the provided properties dictionary
### Version

2.12.0

### Steps to Reproduce

1. Create a PostHogClient.
2. Create a properties dictionary and capture a event
```
var testDictionary = new Dictionary<string, object>();
testDictionary["is_hedgehog_test"] = true;
postHogClient.Capture(
    distinctId: Guid.NewGuid().ToString(),
    eventName: "hedgehog_happy",
    properties: testDictionary,
    groups: null,
    flags: null,
    timestamp: DateTimeOffset.UtcNow
);
```

### Expected Result

The dictionary was not changed by capturing the event.

### Actual Result

The dictionary include `$is_server` and other properties added by capturing.

marandaneto:

`Capture` and several convenience APIs enriched caller-provided property dictionaries in place. Besides the surprising side effect, the batch queue retained the same dictionary, so caller changes after `Capture` could alter the eventual event payload.

The affected core behavior is consumed by `PostHog.AspNetCore` and `PostHog.AI`, so the changeset includes patch releases for all three packages.

Fixes #266.

- `dotnet test -f net8.0 --no-restore --verbosity quiet`
  - 965 unit tests passed, 2 skipped
  - 50 ASP.NET Core tests passed
  - 19 AI tests passed
- `dotnet pack src/PostHog/PostHog.csproj --configuration Release --no-restore --nologo /p:GeneratePackageOnBuild=false`
  - Built `netstandard2.0`, `netstandard2.1`, and `net8.0`
- `bin/fmt --check`
- `pnpm changeset status`
- Bundle-isolated autoreview after the final compatibility fix: no actionable or blocking findings

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent. Read-only scout and reviewer subagents audited public APIs that accept property dictionaries, and the bundle-isolated autoreview helper performed the final review. The implementation uses a shared helper for shallow copies at mutation and ownership boundaries; arbitrary nested mutable values are intentionally not deep-cloned.

### DIFF
--- a/src/PostHog/Api/CapturedEvent.cs
+++ b/src/PostHog/Api/CapturedEvent.cs
@@ -27,7 +27,9 @@ public class CapturedEvent
         DistinctId = distinctId;
         Timestamp = timestamp;
 
-        Properties = properties ?? new Dictionary<string, object>();
+        Properties = properties is null
+            ? new Dictionary<string, object>()
+            : new Dictionary<string, object>(properties);
 
         // Every event has to have these properties.
         Properties[PostHogProperties.DistinctId] = distinctId; // See `get_distinct_id` in PostHog/posthog api/capture.py line 321
--- a/src/PostHog/Api/CapturedEvent.cs
+++ b/src/PostHog/Api/CapturedEvent.cs
@@ -27,9 +27,7 @@ public class CapturedEvent
         DistinctId = distinctId;
         Timestamp = timestamp;
 
-        Properties = properties is null
-            ? new Dictionary<string, object>()
-            : new Dictionary<string, object>(properties);
+        Properties = properties?.Copy() ?? new Dictionary<string, object>();
 
         // Every event has to have these properties.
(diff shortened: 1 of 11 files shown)


src/PostHog/Api/CapturedEvent.cs
```diff
@@ -27,7 +27,9 @@ public CapturedEvent(
         DistinctId = distinctId;
         Timestamp = timestamp;
 
-        Properties = properties ?? new Dictionary<string, object>();
+        Properties = properties is null
+            ? new Dictionary<string, object>()
+            : new Dictionary<string, object>(properties);
```
ioannisj: Repeated a few times, could be abstracted away?

ioannisj (approved):

src/PostHog/Api/CapturedEvent.cs
marandaneto: Good call — addressed in b51935f by adding a shared shallow `Copy()` dictionary helper and using it across all affected ownership boundaries.

PR #268, opened 2026-07-31
feat: support starts_with and ends_with operators in local evaluation

haacked:

PostHog/posthog#72992 added four case-insensitive property filter operators — `starts_with`, `not_starts_with`, `ends_with`, `not_ends_with` — across HogQL, the flags service, and the UI. The UI surface is gated behind a feature flag until server-side SDKs can evaluate these operators locally; on SDK versions without support, flags using them fall back to remote evaluation.

This adds local evaluation support for the four operators, mirroring `icontains`: both sides are stringified and lowercased, and the `not_` variants are exact negations. Filter values are single strings — server-side validation rejects list values for these operators, matching the flags service behavior.

New `ComparisonOperator` members are appended after the existing ones so shipped enum members keep their numeric identity, and the additions are recorded in `PublicAPI.Unshipped.txt` along with the new `PropertyFilterValue.IsPrefixOf`/`IsSuffixOf` methods (modeled on `IsContainedBy`).

New unit tests mirror the flags service's own test matrix for these operators: case-insensitive anchored matching, mid-string non-matches (`prevalue` does not match `starts_with: "Val"`), numeric stringification (`323` matches `starts_with: "3"`, `123` does not), negation inverses, and missing-key inconclusive behavior.

`dotnet test tests/UnitTests --filter "FullyQualifiedName~LocalEvaluatorTests"` passes locally on net8.0 (318 tests, including the 23 new inline cases); the netcoreapp3.1 target needs an x64 runtime not available on this arm64 machine, so CI covers it. `dotnet build src/PostHog` compiles all target frameworks and `dotnet format --verify-no-changes` is clean.

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset file (minor, `PostHog` package)

## 🤖 Agent context

Implemented with PostHog Code as part of rolling out local evaluation for these operators across all server-side SDKs. The implementation deliberately mirrors the existing `icontains` branch in this repo (same string coercion and case folding) and the reference behavior in the Rust flags service (`rust/feature-flags/src/properties/property_matching.rs` in PostHog/posthog). The changeset file was written by hand following the repo's existing format.

---

### DIFF
--- a/src/PostHog/Api/ComparisonOperator.cs
+++ b/src/PostHog/Api/ComparisonOperator.cs
@@ -6,7 +6,7 @@ namespace PostHog.Api;
 /// <summary>
 /// An enumeration representing the comparison types that can be used in a filter.
 /// </summary>
-[JsonConverter(typeof(JsonStringEnumMemberNameJsonConverter<ComparisonOperator>))]
+[JsonConverter(typeof(ComparisonOperatorJsonConverter))]
 public enum ComparisonOperator
 {
     /// <summary>
@@ -181,5 +181,11 @@ public enum ComparisonOperator
     /// Matches if the value does not end with the filter value, ignoring case differences.
     /// </summary>
     [JsonStringEnumMemberName("not_ends_with")]
-    NotEndsWith
+    NotEndsWith,
+
+    /// <summary>
+    /// An operator this version of the SDK doesn't recognize. Flags using such an operator can't be evaluated
+    /// locally and fall back to remote evaluation.
+    /// </summary>
+    Unknown
 }
--- a/src/PostHog/Api/ComparisonOperator.cs
+++ b/src/PostHog/Api/ComparisonOperator.cs
@@ -157,5 +157,29 @@ public enum ComparisonOperator
     /// Matches if the version matches the wildcard pattern (e.g., "1.2.*" means >=1.2.0 and &lt;1.3.0).
     /// </summary>
     [JsonStringEnumMemberName("semver_wildcard")]
-    SemverWildcard
+    SemverWildcard,
+
+    /// <summary>
+    /// Matches if the value starts with the filter value, ignoring case differences.
+    /// </summary>
+    [JsonStringEnumMemberName("starts_with")]
+    StartsWith,
+
+    /// <summary>
+    /// Matches if the value does not start with the filter value, ignoring case differences.
+    /// </summary>
+    [JsonStringEnumMemberName("not_starts_with")]
+    NotStartsWith,
+
+    /// <summary>
+    /// Matches if the value ends with the filter value, ignoring case differences.
+    /// </summary>
+    [JsonStringEnumMemberName("ends_with")]
+    EndsWith,
+
+    /// <summary>
+    /// Matches if the value does not end with the filter value, ignoring case differences.
+    /// </summary>
+    [JsonStringEnumMemberName("not_ends_with")]
+    NotEndsWith
 }
--- a/src/PostHog/Json/PropertyFilterValue.cs
+++ b/src/PostHog/Json/PropertyFilterValue.cs
@@ -140,6 +140,28 @@ public class PropertyFilterValue
         && comparandString.Contains(StringValue, stringComparison);
 
     /// <summary>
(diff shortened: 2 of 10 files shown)


marandaneto (approved):

src/PostHog/Json/PropertyFilterValue.cs
```diff
@@ -131,6 +131,28 @@ public bool IsContainedBy(object? other, StringComparison stringComparison) =>
...
+
+    /// <summary>
+    /// Returns a value indicating whether this instance is a suffix of the specified <paramref name="other"/> instance.
+    /// </summary>
+    /// <param name="other">The other value to compare to this one.</param>
+    /// <param name="stringComparison">The type of comparison if these are strings.</param>
+    /// <returns><c>true</c> if the other value ends with this instance.</returns>
+    public bool IsSuffixOf(object? other, StringComparison stringComparison) =>
+        other?.ToString() is { } comparandString
+        && StringValue is not null
```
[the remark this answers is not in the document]
marandaneto: Confirmed during the cross-SDK review. I agree this is a real but low-frequency consistency risk: decimal stringification can vary with `CurrentCulture`, while server/other-SDK evaluation is locale-independent. It is also inherited from the existing `IsContainedBy` path, so I would not block this PR on it, but an invariant-culture conversion plus a comma-decimal regression test would be the safest resolution (or explicitly defer it as shared cleanup).

src/PostHog/Api/ComparisonOperator.cs
```diff
@@ -157,5 +157,29 @@ public enum ComparisonOperator
     /// Matches if the version matches the wildcard pattern (e.g., "1.2.*" means >=1.2.0 and &lt;1.3.0).
     /// </summary>
     [JsonStringEnumMemberName("semver_wildcard")]
-    SemverWildcard
+    SemverWildcard,
+
+    /// <summary>
+    /// Matches if the value starts with the filter value, ignoring case differences.
+    /// </summary>
+    [JsonStringEnumMemberName("starts_with")]
```
marandaneto, line 165: Could we add a focused JSON deserialization test for all four wire names (`starts_with`, `not_starts_with`, `ends_with`, and `not_ends_with`)? The evaluator tests construct `ComparisonOperator` values directly, and the public API snapshot verifies the enum members but not their JSON mappings, so a typo in one of these attributes would otherwise compile and evade the current focused coverage.

marandaneto (commented):
One additional non-blocking coverage note from the cross-SDK review.

src/PostHog/Api/ComparisonOperator.cs
marandaneto: A related forward-compatibility concern: `JsonStringEnumMemberNameJsonConverter.Read` throws `JsonException` for any operator value that this SDK does not yet know. Because `ComparisonOperator` is nested in the `/flags/definitions` response, one future server-side operator can reject the entire response rather than making only the affected flag inconclusive. On an initial load this eventually causes remote fallback, but when an evaluator is already cached the loader preserves it, so the SDK may continue serving stale local definitions. Could we add an unknown-value test such as `"operator": "future_operator"` and consider mapping unknown strings to an `Unknown` value (or retaining the operator as a string), then letting `LocalEvaluator` mark that flag inconclusive and fall back remotely?

src/PostHog/Json/PropertyFilterValue.cs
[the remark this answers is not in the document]
haacked: Agreed, and fixed family-wide rather than just here: #270 routes every property-value stringification in PropertyFilterValue through `Convert.ToString(value, CultureInfo.InvariantCulture)` with de-DE culture tests. Its PR body notes that this PR's IsPrefixOf/IsSuffixOf get the same treatment once both merge.

src/PostHog/Api/ComparisonOperator.cs
haacked: Done in e8592bf. Added deserialization tests for all four wire names and mapped unknown operators to Unknown so a future server-side operator only makes the affected flag inconclusive (falls back to remote) rather than rejecting the whole response. Left the invariant-culture fix as separate shared cleanup since it predates this PR

PR #270, opened 2026-08-04
fix: use invariant culture when stringifying property values for matching

haacked:
## Problem

Local feature flag evaluation stringifies person and group property values with the machine's current culture. On hosts running comma-decimal locales (de-DE, fr-FR, and most of continental Europe), a numeric property like `3.14` renders as `"3,14"`, so `exact`, `icontains`, regex, and string-fallback comparisons silently diverge from the PostHog flags service, which always renders numbers invariantly (`3.14` → `"3.14"`). The same flag with the same inputs can match remotely but fail local evaluation, purely based on server locale.

## Fix

Stringify with `Convert.ToString(value, CultureInfo.InvariantCulture)` at every site in `PropertyFilterValue` where a property value is compared as a string: `IsRegexMatch`, `IsContainedBy`, both branches of `IsExactMatch`, the string fallback in `CompareTo`, the fallback in `CompareBooleanValue`, and `ParseOverrideSemver`. The class already uses `CultureInfo.InvariantCulture` in its own `ToString` override; this applies the same treatment to the comparison paths.

## Tests

New tests set `CultureInfo.CurrentCulture` to `de-DE` (restoring it after) and assert that a `3.14` double/decimal property matches `exact` and `icontains` filters as if invariant — at the `PropertyFilterValue` unit level and end-to-end through `LocalEvaluator`. Without the fix, 9 of the 11 new cases fail; with it, the full unit suite passes.

## Notes

- #268 adds `IsPrefixOf`/`IsSuffixOf` for `starts_with`/`ends_with`; they stringify the same way and need the identical one-line treatment once it merges.
- `TryCompareNumbers` still parses numeric strings with the current culture (`double.TryParse` without a provider), which is the mirror-image problem for ordered comparisons (`"3.14"` can parse as `314` under de-DE grouping rules). That's out of scope here — this PR is stringification only — but worth a follow-up.

---

### DIFF
--- a/.changeset/invariant-culture-stringification.md
+++ b/.changeset/invariant-culture-stringification.md
@@ -2,4 +2,4 @@
 "PostHog": patch
 ---
 
-Use the invariant culture when stringifying property values for feature flag local evaluation. Numeric property values such as `3.14` now stringify as `"3.14"` regardless of the host locale, so `exact`, `icontains`, regex, and comparison operators match the same way the PostHog flags service does on machines using comma-decimal cultures.
+Use the invariant culture when stringifying property values for feature flag local evaluation. Numeric property values such as `3.14` now stringify as `"3.14"` regardless of the host locale, so `exact`, `icontains`, and regex matching behave the same way the PostHog flags service does on machines using comma-decimal cultures.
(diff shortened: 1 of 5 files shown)


dustinbyrne (approved):
👍 👍

PR #295, opened 2026-08-18
test: cover server SDK User-Agent contract

dustinbyrne:

The server SDK wire snapshots validate request JSON, but did not protect the accompanying `User-Agent` header contract. This adds coverage that every snapshotted request sends the expected `posthog-dotnet/<version>` product and runtime comment containing the framework, OS, and process architecture.

The contract compares the product version with the generated `VersionConstants.Version` value rather than a literal version, so release version bumps are covered without causing test churn. Negative tests prove that unexpected product/version values and missing runtime components are rejected.

- `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --filter 'FullyQualifiedName~ServerSdkSnapshotTests.FinalWireSnapshots|FullyQualifiedName~ServerSdkSnapshotTests.UserAgentContractTests' --no-restore` — passed 9/9 tests.
- `dotnet format tests/UnitTests/UnitTests.csproj --verify-no-changes --no-restore` — passed (workspace-load warnings only).
- The equivalent `netcoreapp3.1` focused command built the projects successfully, but test execution is unavailable on this local Apple Silicon host because VSTest requests an x64 `dotnet` host that is not installed. CI remains responsible for executing that target.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

A human directed the User-Agent contract hardening and PR submission. The Pi coding agent inspected the existing test-only diff, reran focused validation, and submitted it without adding release metadata; the independent final review supplied by the human reported no blockers. No shareable agent-session link is available.

### DIFF
--- a/tests/UnitTests/ServerSdkSnapshotTests.cs
+++ b/tests/UnitTests/ServerSdkSnapshotTests.cs
@@ -1,8 +1,11 @@
+using System.Net.Http.Headers;
+using System.Runtime.InteropServices;
 using System.Text.Json;
 using Microsoft.Extensions.DependencyInjection;
 using PostHog;
 using PostHog.Api;
 using PostHog.Json;
+using PostHog.Versioning;
 using UnitTests.Fakes;
 using Xunit.Sdk;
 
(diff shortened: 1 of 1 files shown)


arnohillen (approved):
<small><em>(created by PostHog Desktop)</em></small>

The asserted shape matches production exactly (`PostHogApiClient.cs:49` sends `posthog-dotnet/{Version} ({Framework}; {OS}; {Arch})`, which `HttpClient` parses into precisely the two `ProductInfoHeaderValue` entries the helper checks). The negative tests fail for the intended reason rather than by accident: `2.13.3-unexpected` is a valid HTTP token and `(; os; arch)` a valid comment, so neither throws a `FormatException` that would mask the `XunitException`.

Issue #47, opened 2025-02-11
Unity support


haacked:
See https://github.com/PostHog/posthog/issues/25705.

haacked:
So here's what I think this entails:

1. New project `PostHog.Unity` which will be published as a new package of the same name.
2. `PostHog.Unity` would build on top of `PostHog.csproj` and target netstandard2.1.
3. New sample project in the samples directory. Ideally this would be a dirt simple Unity app/game so we can test out the native Unity support.

What does native Unity support entail? Here's some ideas from the original issue:

> This integration could include:
 
> * Automatic tracking of common Unity events (e.g., scene loads, object instantiation).
> * Unity Editor plugin for easy PostHog configuration.
> * Performance optimizations to minimize impact on game frame rates.
> * Support for offline event caching and batch sending to handle intermittent internet connections common in mobile games.

marandaneto:
@haacked feel free to tag me for review, I've reviewed a couple of unity PRs in the past as well

haacked:
@RishabhCSL @madratzz - Probably the thing that would help the most is a PR that adds a sample Unity app in the [samples](https://github.com/PostHog/posthog-dotnet/tree/main/samples) directory with instructions on how to develop on it.

A sample app will enable me to test the PostHog client against a real Unity app and expose what issues we may have.

zhuchun:
> [@RishabhCSL](https://github.com/RishabhCSL) [@madratzz](https://github.com/madratzz) - Probably the thing that would help the most is a PR that adds a sample Unity app in the [samples](https://github.com/PostHog/posthog-dotnet/tree/main/samples) directory with instructions on how to develop on it.
> 
> A sample app will enable me to test the PostHog client against a real Unity app and expose what issues we may have.

Hi, What kind of sample do you need, say, a sample that demonstrates how we send HTTP requests in real games? Do you have other requirements like a specific Unity Editor version or game platform?

Ananttah:
Could you please consider adding support for Session Replays as well for Unity mobile games?
I was looking to use Posthog for same but couldn't when I got to know about this limitation.

PR #304, opened 2026-08-25
fix(flags): handle empty evaluation key scopes

marandaneto:

The evaluate-flags specification now distinguishes an omitted key scope from an explicitly empty scope. This implements [sdk-specs PR #47](https://github.com/PostHog/sdk-specs/pull/47) for the .NET SDK.

When `FlagKeysToEvaluate` is omitted or null, `EvaluateFlagsAsync` evaluates all available flags as before. When it is explicitly empty, the method returns an empty snapshot without consulting the cache, loading local definitions, or calling `/flags`, while retaining group context for subsequent `$feature_flag_called` events. A non-empty list continues to scope local evaluation, remote evaluation, and the returned snapshot to those keys.

The nullable public API baseline update is intentional. It preserves the difference between an omitted or null scope and an explicitly empty scope.

- `dotnet test tests/UnitTests/UnitTests.csproj -c Release -f net8.0 --filter 'FullyQualifiedName~TheEvaluateFlagsAsyncMethod' --no-restore` - passed 21 tests with 0 failures.
- `dotnet test tests/UnitTests/UnitTests.csproj -c Release -f netcoreapp3.1 --filter 'FullyQualifiedName~TheEvaluateFlagsAsyncMethod' --no-restore` - passed 21 tests with 0 failures.
- `dotnet build src/PostHog/PostHog.csproj -c Release --no-restore` - built all three target frameworks with 0 warnings and 0 errors.
- `bin/fmt --check` - passed.
- `git diff --check` - passed.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

A Pi worker agent followed the PR workflow and used the bundled Pi autoreview skill. Autoreview flagged the change from a non-null default collection to a nullable property as a compatibility concern. The change was kept because the accepted SDK specification requires .NET to preserve the difference between an omitted or null scope and an explicitly empty scope. The public API baseline and XML documentation make that intentional contract visible to reviewers.

### DIFF
--- a/.changeset/empty-evaluation-key-scopes.md
+++ b/.changeset/empty-evaluation-key-scopes.md
@@ -1,5 +1,5 @@
 ---
-"PostHog": patch
+"PostHog": minor
 "PostHog.AspNetCore": patch
 ---
 
--- a/.changeset/empty-evaluation-key-scopes.md
+++ b/.changeset/empty-evaluation-key-scopes.md
@@ -1,0 +1,6 @@
+---
+"PostHog": patch
+"PostHog.AspNetCore": patch
+---
+
+Return an empty feature flag snapshot without evaluation work when `FlagKeysToEvaluate` is explicitly empty.
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -1061,6 +1061,11 @@ public sealed class PostHogClient : IPostHogClient
             return FeatureFlagEvaluations.Empty(_evaluationsHost, resolvedDistinctId ?? string.Empty);
         }
 
+        if (options?.FlagKeysToEvaluate is { Count: 0 })
+        {
+            return FeatureFlagEvaluations.Empty(_evaluationsHost, resolvedDistinctId ?? string.Empty);
+        }
+
         if (RequiresMissingPersonalApiKey(options, nameof(EvaluateFlagsAsync)))
         {
             return FeatureFlagEvaluations.Empty(_evaluationsHost, resolvedDistinctId ?? string.Empty);
--- a/src/PostHog/PostHogClient.cs
+++ b/src/PostHog/PostHogClient.cs
@@ -1063,7 +1063,10 @@ public sealed class PostHogClient : IPostHogClient
 
         if (options?.FlagKeysToEvaluate is { Count: 0 })
         {
-            return FeatureFlagEvaluations.Empty(_evaluationsHost, resolvedDistinctId ?? string.Empty);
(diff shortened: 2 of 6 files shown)


.changeset/empty-evaluation-key-scopes.md
```diff
@@ -0,0 +1,6 @@
+---
+"PostHog": patch
```
turnipdabeets: [question] `FlagKeysToEvaluate` goes from a non-null `[]` default to nullable, so `new AllFeatureFlagsOptions().FlagKeysToEvaluate` returns null where it used to return an empty list — an NRE for anyone reading it back, and a CS8602 warning in nullable-enabled projects. Nothing else can satisfy the spec here, so the change itself looks right to me; should this be `minor` rather than `patch` on 2.9.0?

src/PostHog/PostHogClient.cs
```diff
@@ -1061,6 +1061,11 @@ public async Task<FeatureFlagEvaluations> EvaluateFlagsAsync(
             return FeatureFlagEvaluations.Empty(_evaluationsHost, resolvedDistinctId ?? string.Empty);
         }
 
+        if (options?.FlagKeysToEvaluate is { Count: 0 })
+        {
+            return FeatureFlagEvaluations.Empty(_evaluationsHost, resolvedDistinctId ?? string.Empty);
```
turnipdabeets: [suggestion] `FeatureFlagEvaluations.Empty` hardcodes `groups: null`, and `CaptureFeatureFlagCalled` is handed `groups: _groups` — so accessing a missing key on an empty-scope snapshot fires `$feature_flag_called` without `$groups`, while the same call with a non-empty scope includes them. This is the first time `Empty` runs with a real distinct id, so those events actually fire now. Rust, Python, Ruby, PHP and Node all thread groups into their empty snapshot — could we pass `options?.Groups` through here too?

turnipdabeets (approved):

.changeset/empty-evaluation-key-scopes.md
marandaneto: Good catch! Updated in 94052c2 — the PostHog changeset now requests a minor release for the nullable public API change, while PostHog.AspNetCore remains a patch release for its dependency update.

src/PostHog/PostHogClient.cs
marandaneto: Good catch! Fixed in 94052c2 — explicit-empty snapshots now retain options.Groups, with regression coverage asserting that the resulting $feature_flag_called event includes $groups.

PR #306, opened 2026-08-27
fix(flags): match local string filters like the flags service

### LINKED ISSUE #78019: Case folding and numeric stringification are inconsistent between SDK local evaluation and the flags service
While shipping `starts_with`/`ends_with` local evaluation across the SDKs (#72992), a pattern emerged: our case-insensitive property matching is inconsistent between the SDKs and the flags service, and within the flags service itself. Filing this so we can settle it once instead of per code review.

## How the flags service folds case

In `rust/feature-flags/src/properties/property_matching.rs`:

- `icontains` / `starts_with` / `ends_with` (and negations): `to_ascii_lowercase()`, ASCII-only. The code comment says it's a deliberate perf choice.
- `exact` / `is_not`: full Unicode `to_lowercase()`.

So the server already applies two different folding rules depending on operator.

## What the SDKs use for the icontains/starts_with/ends_with family

| SDK | Folding | Matches server? |
|---|---|---|
| PHP | `strtolower` (ASCII-only) | ✅ |
| Ruby | `downcase(:ascii)` as of PostHog/posthog-ruby#229 | ✅ |
| Elixir | ASCII-only in PostHog/posthog-elixir#192 (pending) | ✅ |
| Python | `casefold()` (most aggressive Unicode; `"ss"` matches `"ß"`) | ❌ |
| Go | `strings.ToLower` (Unicode) | ❌ |
| Node | `toLowerCase()` (Unicode) | ❌ |
| Rust SDK | `to_lowercase()` (Unicode) | ❌ |
| .NET | `OrdinalIgnoreCase` (Unicode simple folding) | ❌ |

Concretely: person property `"Äbc"` with `starts_with: "ä"` matches locally in Python/Go/Node/Rust/.NET and does **not** match on `/flags`. Same flag, same person, different answer depending on evaluation path — exactly the divergence local evaluation is supposed to avoid.

## Adjacent inconsistencies in the same code

- The Rust SDK's `exact` operator uses `eq_ignore_ascii_case` while the server's `exact` is Unicode — the mirror image of the divergence above.
- Numeric stringification: the server renders a float-typed `323.0` as `"323.0"` (serde `Display`), as do Ruby and Python, but Go and Node render `"323"`, so a JSON `323.0` property matches `ends_with: "3"` locally but not remotely. .NET additionally used the host locale for decimal separators (`3.14` → `"3,14"` on de-DE), fixed in PostHog/posthog-dotnet#270.

## Options

1. **Standardize the SDKs on ASCII-only folding** for the icontains/starts/ends family to match the server. PHP and Ruby are already there; the rest are one-line changes plus tests. The numeric stringification fixes for Go/Node belong in the same sweep.
2. **Make the server Unicode-aware** and leave the SDKs alone. Friendlier for international text, but changes existing server behavior, and the ASCII choice was deliberate.
3. **Leave it and document** the edge case.

I lean toward 1, plus an explicit decision on whether the server's Unicode `exact`/`is_not` is intentional. Non-ASCII values differing only by case are rare, but today the answer depends on which SDK and which operator you hit, and that's the worst version.

marandaneto:

Local feature flag evaluation must return the same result as the currently released Rust flags service for the same property values. The previous implementation used .NET's general case-insensitive comparison and host `ToString()` behavior, which diverged from `/flags`. This fixes PostHog/posthog#78019.

This PR intentionally follows the **current released backend**, not the alternative semantics discussed in draft backend PR PostHog/posthog#90694.

The evaluator now mirrors the backend by:

- applying the boolean gate before ordinary list membership, including recursively boolean-like arrays, aggregate truthiness, vacuous truth for `[]`, and complete `is_not` complementation;
- stringifying representable JSON values like `serde_json`, including compact arrays, UTF-8-key-sorted objects, `null`, negative zero, integral-float preservation, and the backend's exponent cutovers;
- using full Unicode lowercase behavior for `exact`/`is_not`, including dotted-I expansion and contextual `Final_Sigma` handling for scalar and list filters;
- retaining ASCII-only lowercase behavior for contains, starts-with, ends-with, and their negations.

The contextual sigma helper uses checked-in generated Unicode 17.0.0 `Cased` and `Case_Ignorable` ranges. The generator pins the official UCD source SHA-256 hashes, validates range counts/order, carries Unicode provenance/license links, and reproduces the checked-in source byte-for-byte. Cyclic, excessively deep, or otherwise non-JSON CLR collections fail matching safely rather than risking recursion crashes or partial JSON output.

A patch changeset covers `PostHog` and `PostHog.AspNetCore`.

Red-green regression tests cover boolean scalar/list precedence, empty and nested arrays, `is_not`, canonical JSON objects/arrays/null/floats, final and medial sigma contexts, dotted I, scalar/list filters, and unsafe CLR collection graphs.

- `bin/generate-unicode-lowercase-data` plus byte-for-byte comparison with the checked-in generated file
- SHA-256 verification of Unicode 17.0.0 `DerivedCoreProperties.txt` and `SpecialCasing.txt`
- 80-value finite `f64` differential stringification check against `serde_json`
- `bin/fmt --check`
- `dotnet build --configuration Release --no-restore --nologo` (0 warnings, 0 errors)
- `dotnet test tests/UnitTests/UnitTests.csproj --framework net8.0 --configuration Release --no-build --nologo` (1,137 passed, 2 skipped)
- `dotnet test tests/UnitTests/UnitTests.csproj --framework netcoreapp3.1 --configuration Release --no-build --nologo` (1,129 passed, 2 skipped)
- Bundled autoreview against the dirty local change bundle (clean after addressing collection-safety findings)

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

A human directed the cross-SDK parity work for PostHog/posthog#78019. Pi reproduced the reported differences against the current Rust backend, implemented the .NET parity fixes test-first, and ran the bundled isolated autoreview. The generated Unicode data-table approach was explicitly approved for exact `Final_Sigma` context handling without adding a runtime dependency.


src/PostHog/Json/PropertyFilterValue.cs
```diff
@@ -222,9 +238,74 @@ or TypeCode.SByte or TypeCode.UInt16 or TypeCode.UInt32 or TypeCode.UInt64
...
+            : stringValue;
+    }
+
+#pragma warning disable CA1308 // The flags service lowercases both operands; uppercasing has different Unicode semantics.
+    static bool UnicodeLowercaseEquals(string left, string? right) =>
+        right is not null
+        && string.Equals(UnicodeLowercase(left), UnicodeLowercase(right), StringComparison.Ordinal);
+
+    // .NET's invariant lowercase mapping omits the full Unicode expansion for LATIN CAPITAL LETTER I WITH DOT ABOVE.
+    static string UnicodeLowercase(string value)
```
ioannisj: [suggestion] `UnicodeLowercase` handles U+0130 but not Final_Sigma, which Rust's `to_lowercase()` does apply.

`"ΟΔΟΣ"` comes out as `οδος` (final sigma, U+03C2) in Rust and `οδοσ` (U+03C3) here, so an `Exact` filter on `"ΟΔΟΣ"` matches a person property of `"οδος"` server side and misses locally.

I swept every code point comparing the two mappings and U+0130 really is the only per-character difference, so this is close. Final_Sigma is contextual rather than per-char, which is why the tests don't catch it: a lone `Σ` has no preceding cased char, so Rust gives `σ` too.

A post-pass after `ToLowerInvariant` mapping U+03A3 to U+03C2 when it's preceded by a cased char and not followed by one would close it. It's Greek only, so documenting the gap instead is fine by me. Worth a line in the changeset either way?

src/PostHog/Json/PropertyFilterValue.cs
```diff
@@ -161,6 +162,21 @@ public bool IsSuffixOf(object? other, StringComparison stringComparison) =>
         && StringValue is not null
         && comparandString.EndsWith(StringValue, stringComparison);
 
+    internal bool IsContainedByAsciiIgnoreCase(object? other) =>
+        ToInvariantString(other) is { } comparandString
+        && StringValue is not null
+        && ToAsciiLowercase(comparandString).Contains(ToAsciiLowercase(StringValue), StringComparison.Ordinal);
+
+    internal bool IsPrefixOfAsciiIgnoreCase(object? other) =>
```
ioannisj: [suggestion] Worth deleting `IsPrefixOf`/`IsSuffixOf` in this PR? They came in with #268 and are still in `PublicAPI.Unshipped.txt`, so they've never been released, and now that these Ascii variants take over the only dispatch site in `LocalEvaluator` they have zero callers and zero tests.

This is the free window. Otherwise we ship a `StringComparison`-based surface that doesn't reflect how `StartsWith`/`EndsWith` actually evaluate, and then we're stuck with it.

`IsContainedBy` is shipped so it has to stay, but its remarks could say it isn't the flag evaluation path, since the test at `tests/UnitTests/Json/PropertyFilterValueTests.cs:118` currently reads like it documents current semantics.

src/PostHog/Json/PropertyFilterValue.cs
```diff
@@ -194,7 +210,7 @@ static bool IsExactListMatch(
         }
 
         var stringValue = ToInvariantString(overrideValue);
-        if (stringValue is not null && values.Contains(stringValue, StringComparer.OrdinalIgnoreCase))
+        if (stringValue is not null && values.Any(value => UnicodeLowercaseEquals(value, stringValue)))
```
ioannisj: [nit] `UnicodeLowercase(stringValue)` runs once per element here, a `StringBuilder` plus two strings each time even when there's no U+0130 anywhere. Hoisting the lowered override out of the lambda makes it one allocation instead of N.

Cheap extra while you're in there: fast-path `UnicodeLowercase` to plain `ToLowerInvariant()` when `value.IndexOf('\u0130') < 0`, which is basically always.

ioannisj (approved):

arnohillen:
`UnicodeLowercase` at `src/PostHog/Json/PropertyFilterValue.cs:293` ends in `ToLowerInvariant()`, which has no Final_Sigma mapping, so `exact` with filter `ΠΑΡΑΓΓΕΛΙΕΣ` against property `παραγγελιες` returns `false` here (folds to `...U+03C3`) where the service's `to_lowercase` in `rust/feature-flags/src/properties/property_matching.rs:230` returns `true` (folds to `...U+03C2`), a regression from `main`'s `OrdinalIgnoreCase` which returned `true`; this also hits `is_not` and the `is_in` list path via `IsExactListMatch`, and the added `("ς", "Σ", false)` case stays correct under a Final_Sigma-aware fold.

src/PostHog/Json/PropertyFilterValue.cs
```diff
@@ -222,9 +238,74 @@ or TypeCode.SByte or TypeCode.UInt16 or TypeCode.UInt32 or TypeCode.UInt64
     }
 
     // Override values must stringify with the invariant culture ("3.14", never "3,14") to match how the
-    // PostHog flags service stringifies values. Null stays null so null overrides never match string filters.
-    static string? ToInvariantString(object? value) =>
-        value is null ? null : Convert.ToString(value, CultureInfo.InvariantCulture);
+    // PostHog flags service stringifies values. Preserve the decimal point of integral floating-point values,
+    // because the service's JSON representation distinguishes 323.0 from the integer 323.
+    static string? ToInvariantString(object? value) => value switch
```
haacked: `blocking`: A whole-number `double` or `float` person property now evaluates one way locally and the other way at `/flags`, so the same call gives a different flag result depending on which path handled it.

The new floating-point branch converts a C# `double` of `323.0` to the string `323.0`. That same value reaches the service as the JSON number `323`, because System.Text.Json writes the shortest round-trip form and `JsonSerializerHelper.Options` sets no `NumberHandling`. Serializing a person-properties dictionary through the SDK's own options gives `{"person_properties":{"rev":323}}`. The service stringifies that as `323` and matches a filter of `"323"`, while local evaluation produces `323.0` and misses. Before this PR, both sides produced `323`.

The same dictionary reaches both paths: `PostHogClient` hands `options.PersonProperties` to local evaluation and the identical object to `GetFeatureFlagsAsync`, and the service prefers request properties over stored ones. The impact is wider than exact matching. `ToInvariantString` also feeds `IsRegexMatch`, the three new ASCII contains/starts-with/ends-with helpers, and the list path. A filter of `^323$` stops matching a `323.0` property too.

`decimal` already works without any special case: System.Text.Json writes `323.0m` as `323.0`, and `Convert.ToString(323.0m, CultureInfo.InvariantCulture)` returns `"323.0"`.

Dropping the branch leaves `ToInvariantString` as `value is null ? null : Convert.ToString(value, CultureInfo.InvariantCulture)` and removes both `StringifyFloatingPoint` overloads. Four test rows flip with it: [`PropertyFilterValueTests.cs:78`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Json/PropertyFilterValueTests.cs#L78) and [`:79`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Json/PropertyFilterValueTests.cs#L79), and [`LocalEvaluatorTests.cs:417`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Features/LocalEvaluatorTests.cs#L417) and [`:431`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Features/LocalEvaluatorTests.cs#L431). The two integer rows at [`PropertyFilterValueTests.cs:80`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Json/PropertyFilterValueTests.cs#L80) and [`:81`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Json/PropertyFilterValueTests.cs#L81) keep their current expectations.

[`LocalEvaluatorTests.cs:431`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Features/LocalEvaluatorTests.cs#L431) is worth reading on its own. It asserts that `323.0` does not end with `"3"`, while the service, handed the `323` the SDK actually sends, says it does.

If the branch stays, the two `StringifyFloatingPoint` overloads differ only in which type's `IsNaN` and `IsInfinity` they call and would collapse into one.

src/PostHog/Json/PropertyFilterValue.cs
haacked: `blocking`: A word-final Greek `Σ` folds to `σ` here and to `ς` at the service, so `exact` on Greek text disagrees with `/flags`. @arnohillen raised this already, and it reproduces exactly as described. Two things to add to it.

U+03A3 is the only character where this helper disagrees with the service. Comparing `UnicodeLowercase` against Rust's `to_lowercase` across every code point turned up no divergence, and fuzzing one-to-three character strings over a Greek and Latin alphabet with combining marks produced 330 divergences, all of them containing `Σ`. Handling it makes this helper agree with the service everywhere, not just fixing one more symptom.

The miss also runs in the other direction. A property spelled with a trailing `σ` now matches an uppercase filter where the service says it should not, so the PR is wrong both ways here, where `main` was wrong one way.

A fold that covers it, checked against `Σ` alone, `ΑΣΒ`, and `ΟΔΟΣ.`, and reported to keep the four new expectations at [`PropertyFilterValueTests.cs:13-16`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Json/PropertyFilterValueTests.cs#L13-L16) green:

```csharp
static string UnicodeLowercase(string value)
{
    var expanded = new StringBuilder(value.Length + 1);
    for (var index = 0; index < value.Length; index++)
    {
        var character = value[index];
        if (character == '\u0130')
        {
            expanded.Append("i\u0307");
        }
        else if (character == '\u03A3')
        {
            var wordFinal = CasedAfterIgnorable(value, index - 1, -1)
                            && !CasedAfterIgnorable(value, index + 1, 1);
            expanded.Append(wordFinal ? '\u03C2' : '\u03C3');
        }
        else
        {
            expanded.Append(character);
        }
    }
    return expanded.ToString().ToLowerInvariant();
}

// Unicode Final_Sigma: skip case-ignorable characters, then report whether the next one is cased.
static bool CasedAfterIgnorable(string value, int start, int step)
{
    for (var index = start; index >= 0 && index < value.Length; index += step)
    {
        var character = value[index];
        if (char.GetUnicodeCategory(character) is UnicodeCategory.NonSpacingMark
            or UnicodeCategory.EnclosingMark or UnicodeCategory.Format
            or UnicodeCategory.ModifierLetter or UnicodeCategory.ModifierSymbol)
        {
            continue;
        }
        return char.IsUpper(character) || char.IsLower(character)
               || char.GetUnicodeCategory(character) == UnicodeCategory.TitlecaseLetter;
    }
    return false;
}
```

`("παραγγελιες", "\"ΠΑΡΑΓΓΕΛΙΕΣ\"", true)` would be the regression case to add alongside it.

src/PostHog/Json/PropertyFilterValue.cs
haacked: `suggestion`: No test separates this line's Unicode fold from the comparer it replaced, so the old `StringComparer.OrdinalIgnoreCase` could go back in here with every test still green, silently restoring the old fold for list-valued exact filters. Every list case in the theory above is ASCII, where both give the same result. A filter of `["Σ"]` against a property of `"ς"` expecting false would separate them, since `StringComparer.OrdinalIgnoreCase.Equals("Σ", "ς")` returns true.

Separately, the override value is folded once per list element here, because `stringValue` goes into the lambda and `UnicodeLowercaseEquals` folds both arguments on every call. Each fold allocates a `StringBuilder`, its buffer, and an intermediate string, so a 500-value allowlist does about twice the work it needs to. Hoisting the override's fold out of the loop gives the same answer, since it produces the same string every iteration:

```csharp
var stringValue = ToInvariantString(overrideValue);
var loweredOverride = stringValue is null ? null : UnicodeLowercase(stringValue);
if (loweredOverride is not null
    && values.Any(value => string.Equals(UnicodeLowercase(value), loweredOverride, StringComparison.Ordinal)))
{
    return true;
}
```

src/PostHog/Json/PropertyFilterValue.cs
```diff
@@ -222,9 +238,74 @@ or TypeCode.SByte or TypeCode.UInt16 or TypeCode.UInt32 or TypeCode.UInt64
...
+            else
+            {
+                expanded.Append(character);
+            }
+        }
+        return expanded.ToString().ToLowerInvariant();
+    }
+#pragma warning restore CA1308
+
+    static string ToAsciiLowercase(string value)
```
haacked: `suggestion`: Two case-folding helpers sit next to each other with no explanation of why they differ. Someone tidying this file could unify them and accidentally undo the parity this PR establishes. `ToAsciiLowercase` folds only `A-Z`, while [`UnicodeLowercase` on line 279](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/src/PostHog/Json/PropertyFilterValue.cs#L279) folds all of Unicode.

The difference is surprising enough to merit a comment: `Äbc` doesn't start with `ä` under contains/starts-with/ends-with, but `ä` equals `Ä` under exact matching. Both are asserted by tests this PR adds, so the cleanup would turn them red without explaining why they were there.

```suggestion
    // The flags service folds only A-Z for contains, starts-with, and ends-with, so `Äbc` does not
    // start with `ä`. Exact matching goes through UnicodeLowercase instead and does fold `Ä` to `ä`.
    static string ToAsciiLowercase(string value)
```

Omit the Rust comment about ASCII folding being faster. Here it's a parity requirement, not a performance optimization, so a maintainer reading it as a speed choice might feel free to change it.

.changeset/case-folding-parity.md
```diff
@@ -0,0 +1,6 @@
+---
+"PostHog": patch
+"PostHog.AspNetCore": patch
+---
+
+Match local feature flag string operators using the same ASCII and Unicode lowercase rules as the flags service.
```
haacked: `suggestion`: The changeset doesn't explain which flags can change value after upgrading. Anyone with a non-ASCII person property matched by contains, starts-with, or ends-with will see those comparisons start returning false, and any flag rolling out to those people stops rolling out locally. The new test case at [`LocalEvaluatorTests.cs:375`](https://github.com/PostHog/posthog-dotnet/blob/1a4c22d43edf27211964b14ce1030b0dfb1f1fde/tests/UnitTests/Features/LocalEvaluatorTests.cs#L375) shows this: `Äbc` no longer contains `ä`.

`patch` is the right level since this fixes a bug rather than adding an operator. The issue is just the wording.

```suggestion
Local feature flag evaluation now folds case the same way the flags service does, so a flag can evaluate differently after this upgrade.

`icontains`, `not_icontains`, `starts_with`, `not_starts_with`, `ends_with`, and `not_ends_with` fold only ASCII `A-Z`. A filter of `ä` no longer matches a property value of `Äbc` locally. The flags service already behaved this way, so remote evaluation is unchanged and local results now agree with it.

`exact` and `is_not` fold the full Unicode range, including the expansion of `İ` to `i` plus a combining dot.
```

marandaneto:
@arnohillen Addressed in 1e4cf2130410b8c124082dff72a7cbdd70acd134. You were right that the previous helper did not reproduce Rust’s contextual `Final_Sigma` result.

The fix now applies the Unicode 17.0.0 `Final_Sigma` condition using the official `Cased` and `Case_Ignorable` derived-property ranges before invariant lowercase, while retaining the `İ -> i + U+0307` expansion. The checked-in generated table is reproducible from pinned UCD files (SHA-256 `24c7fed…c8` and `efc25faf…88`) and includes Unicode provenance/license links.

Regression coverage now exercises `ΠΑΡΑΓΓΕΛΙΕΣ`, `ΟΔΟΣ`, final versus medial sigma (including case-ignorable context), dotted I, scalar and list filters, and both `exact` and `is_not`. The final Release suites pass on both targets: net8.0 1,137 passed / 2 skipped; netcoreapp3.1 1,129 passed / 2 skipped.

marandaneto:
A note on why this PR includes `bin/generate-unicode-lowercase-data` and checked-in `UnicodeSpecialCasingData.cs` instead of using a built-in .NET comparison:

The flags service performs Rust full Unicode lowercase followed by ordinal equality. No .NET built-in reproduces that exact operation:

- `"ΟΣ".ToLowerInvariant()` produces `"οσ"`, while Rust/backend lowercase produces `"ος"` because it applies the contextual `Final_Sigma` rule.
- `"İ".ToLowerInvariant()` remains `"İ"`, while Rust/backend lowercase expands it to `"i\u0307"`.
- `StringComparison.OrdinalIgnoreCase` is not equivalent: it considers `"ΟΣ"` equal to both `"ος"` and `"οσ"`, while the backend lowercase-then-ordinal behavior matches only `"ος"`.
- `TextInfo.ToLower`, Greek culture casing, and `Rune.ToLowerInvariant` do not provide the required contextual and expanding full-lowercase mapping either.

The generator is maintenance-only and is not invoked by SDK consumers at runtime. It produces a checked-in, deterministic table containing only the Unicode `Cased` and `Case_Ignorable` ranges required to evaluate `Final_Sigma`; dotted-I expansion remains a direct mapping. This keeps behavior stable across `netstandard2.0`, `netstandard2.1`, and `net8.0` without adding a considerably heavier full Unicode dependency such as ICU4N.

PR #312, opened 2026-09-02
fix(flags): handle null early_exit values

### LINKED ISSUE #311: EarlyExit not being nullable is causing local eval to blowup
### Version

2.9.4

### Steps to Reproduce

Set local eval to true when setting up your CustomFeatureProvider : PostHogFeatureFlagContextProvider class
Create a new feature flag through posthog ui
Make a request to get feature flags within a running app
View logs spitting out an exception like:

Exception

[FEATURE FLAGS] Unable to get feature flags and payloads
EventId: 1002 (LogErrorUnableToGetFeatureFlagsAndPayloads)
SourceContext: PostHog.Api.PostHogApiClient

System.Text.Json.JsonException: The JSON value could not be converted to System.Boolean.
Path: $[6].filters.early_exit | LineNumber: 0 | BytePositionInLine: 5641.
 ---> System.InvalidOperationException: Cannot get the value of a token type 'Null' as a boolean.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_ExpectedBoolean(JsonTokenType tokenType)
   ...
   at PostHog.Api.PostHogApiClient.GetAuthenticatedResponseWithETagAsync[T](String relativeUrl, String etag, CancellationToken cancellationToken)
   at PostHog.Api.PostHogApiClient.GetFeatureFlagsForLocalEvaluationAsync(String etag, CancellationToken cancellationToken)

### Expected Result

It should not blow up when a feature flag with a null EarlyExit property is returned when trying to evaluate flags locally.

### Actual Result

It blows up

dustinbyrne:

Fixes #311.

The `/flags/definitions` response can contain `filters.early_exit: null`. The SDK modeled this field as a non-nullable boolean, so `System.Text.Json` rejected the response before local feature flag evaluation could begin.

This makes the internal field nullable while preserving `false` as the absent-field default. The evaluator already normalizes null to `false`, so explicit null values now follow the existing non-early-exit path without changing true, false, or omitted-field behavior.

A patch changeset covers `PostHog`. There are no public API or default-behavior changes.

- Reproduced the issue with a regression case that failed before the fix with `JsonException: The JSON value could not be converted to System.Boolean`.
- Extended the existing JSON-backed early-exit test to verify both `true` and `null`, including the resulting local evaluation behavior.
- `dotnet restore --locked-mode`
- `bin/fmt --check`
- `dotnet build --configuration Release --no-restore --nologo` (0 warnings, 0 errors)
- `dotnet test --configuration Release --no-build --nologo --framework net8.0`
  - `UnitTests`: 1,141 passed, 2 skipped
  - `UnitTests.AspNetCore`: 50 passed
  - `PostHog.AI.Tests`: 19 passed

The .NET Core 3.1 test binary builds successfully, but its tests were not run locally because the x64 .NET Core 3.1 host is not installed in this environment.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi coding agent (`openai-codex/gpt-5.6-sol`) traced and reproduced the reported serialization failure, implemented the scoped fix test-first, and ran the Simplify skill. A fresh-context builtin reviewer found no blockers and suggested exercising evaluation after deserialization; the final regression theory covers that path for both `true` and `null`.

### DIFF
--- a/.changeset/pink-items-stick.md
+++ b/.changeset/pink-items-stick.md
@@ -1,5 +1,6 @@
 ---
 "PostHog": patch
+"PostHog.AspNetCore": patch
 ---
 
 Allow local feature flag definitions with a null `early_exit` value.
--- a/.changeset/pink-items-stick.md
+++ b/.changeset/pink-items-stick.md
@@ -1,0 +1,5 @@
+---
+"PostHog": patch
+---
+
+Allow local feature flag definitions with a null `early_exit` value.
--- a/src/PostHog/Api/LocalEvaluationApiResult.cs
+++ b/src/PostHog/Api/LocalEvaluationApiResult.cs
@@ -167,10 +167,11 @@ internal record FeatureFlagFilters
     /// When <c>true</c>, local condition evaluation stops and returns a definitive disabled result
     /// as soon as a condition group's property filters match (or the group has no property filters)
     /// but the rollout percentage excludes the user, instead of falling through to later condition
-    /// groups. Mirrors the server-side Rust evaluation engine. Defaults to <c>false</c> when absent.
+    /// groups. Mirrors the server-side Rust evaluation engine. Defaults to <c>false</c> when absent;
+    /// explicit <c>null</c> values are also treated as <c>false</c> during evaluation.
     /// </summary>
     [JsonPropertyName("early_exit")]
-    public bool EarlyExit { get; init; }
+    public bool? EarlyExit { get; init; } = false;
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


.changeset/pink-items-stick.md
```diff
@@ -0,0 +1,5 @@
+---
+"PostHog": patch
```
marandaneto, line 2: also add aspnet

marandaneto (approved):

PR #319, opened 2026-09-09
fix: omit null object properties from serialized events

marandaneto:

Event properties currently retain null-valued object members on the wire. This implements the recursive omission rule in [sdk-specs PR #60](https://github.com/PostHog/sdk-specs/pull/60) without narrowing supported public inputs.

The core event serializer and immediate capture boundary now omit those members after enrichment and hooks. Arrays keep their null positions. Empty objects, other scalar values, ordered duplicate and case-distinct property names, and caller-owned data are preserved. Generic JSON serialization for flags and caches is unchanged. Only the root `$exception_list` on `$exception` keeps its existing typed metadata semantics.

AI events use the same core serializer, with no AI-specific cleanup. A patch changeset selects only `PostHog`, where the production change lives. No versions or dependencies are changed.

- 35 selected core tests passed, including actual client capture, exception capture, before-send drops/enrichment, intercepted plain/gzip batch bytes and immediate Identify HTTP contents.
- Eight boundary cases passed against the netstandard2.1 library through the legacy test target. Coverage includes custom converters, POCO/DOM inputs, null array positions, duplicate keys, numeric token precision, generic serialization and caller nonmutation.
- One actual nonstreaming OpenAI handler/context test passed through the core HTTP-content boundary. Both provider responses and SDK delivery use in-process fake transports. No live ingestion or provider requests were made.
- Core netstandard2.0, netstandard2.1 and net8.0 builds passed with public API analyzers enabled and zero warnings/errors. Changed-file whitespace/style checks passed, with two formatter workspace-loading warnings.
- Isolated committed-branch autoreview passed at `1886b0f5d701982624d5a2fd57111318a2d7ada8` against `eadb15eaa78ee86469f9841eb23b8b14770b9e96`, with no findings.
- All 19 registered CI checks passed on that HEAD, including build/package, Test, Format, SDK compliance, CodeQL, Semgrep and Wiz checks. The existing SDK compliance harness is not evidence of the new null-property contract.

**Validation limits:** Tests ran on .NET 9.0.19 with major-version roll-forward, not actual .NET 3.1/8 runtimes. netstandard2.0 has build-only evidence. The preexisting ReadOnlyDictionary converter stack overflow remains deferred and is not newly declared unsupported. The full suite, minimum-runtime matrix, AI streaming/error paths, retries, every immediate API and performance were not validated locally. This SDK has no disk-backed event queue. These gaps remain open for human review and are not cleared merge or release gates.

## :pencil: Checklist

- [x] I updated the docs if needed. The shared specification is linked above.

### If releasing new changes

A core-only patch changeset was written directly. The package-manager command was not run.

## 🤖 Agent context

Pi agents assisted with implementation, focused validation and review using Git, the .NET CLI, the isolated autoreview helper and GitHub CLI. @marandaneto directed the work and is the DRI. Session details remain private.

The change stays in the shared core serializer rather than duplicating policy in AI. Ordered JsonElement traversal replaced an earlier mutable-object approach that rejected supported duplicate/case-distinct keys. This is a draft for human review, not release approval.


marandaneto:
gonna be done server side anyway

PR #330, opened 2026-09-21
ci: pin conventional-commit titles for dependabot PRs

turnipdabeets:
## Problem

Dependabot picks its PR title style from recent commit history. Most of its PRs here come out as `chore(deps): …`, but #328 came out as `Bump the nuget-dependencies group with 18 updates`, which failed the Conventional Commits title check and had to be retitled by hand.

## Changes

Set `commit-message: { prefix: "chore", include: "scope" }` for all three ecosystems, so titles are always `chore(deps): …` / `chore(deps-dev): …` — the same form they already take when the inference works.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marandaneto:
noice

marandaneto (approved):

